PR #6, opened 2021-10-19
Remove unused Table of Contents depedencies from front-end

### LINKED ISSUE #5: Table of Contents enqueues unnecessary assets on front-end
cata-toc-script requires wp-polyfill and wp-i18n

wp-polyfill requires regenerator-runtime
wp-i18n requirs wp-hooks

Screen Shot 2021-10-19 at 9 47 16 AM

douglas-johnson:
### Related Issues

- Resolves #5 

### What Was Accomplished

- Updated `@wordpress/scripts` locally and globally.
- Added custom Webpack config for the table of contents block.
  - It sets `injectPolyfill:false` in `DependencyExtractionWebpackPlugin`, which removes the `wp-polyfill` dependency. It can be added separately for the block editor. Many other blocks require it explicitly so it is likely to be there. We don't need it on the front-end since we didn't do anything polyfillable in the TOC script.
  - This has also been addressed in https://github.com/WordPress/gutenberg/pull/35436 which isn't available on NPM yet.
- Updated the `build:table-of-contents` script to use the updated config.
- Removed use of `wp-i18n` from the editor scripts.
- Removed the textdomain from block.json which stops of the `wp-i18n` dependency on the front-end.

### How It Was Tested

- [x] Local ( see screenshot below )
- [x] Staging

### Screenshots
Screen Shot 2021-10-19 at 3 04 53 PM

### DIFF
--- a/cata-blocks.php
+++ b/cata-blocks.php
@@ -12,7 +12,7 @@
  * Description: Block Editor components for use with the Cata theme.
  * Author:      Thought & Expression Co. <devjobs@thought.is>
  * Author URI:  https://thought.is
- * Version:     0.3.4-beta1
+ * Version:     0.3.4-beta2
  * License:     GPL v3 or later
  * License URI: http://www.gnu.org/licenses/gpl-3.0.txt
  */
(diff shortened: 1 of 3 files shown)


richdacuban (approved):
Tested Locally. Worked as expected!

PR #7, opened 2021-11-05
newsletter block, no front end styling

richdacuban:
### Related issue(s)
- related to [CC newsletter design update](https://github.com/thoughtis/wpe-creepycatalog/issues/227)

### Accomplishes
- Creates a newsletter signup form block which can be used in child themes
Screen Shot 2021-11-05 at 11 33 13 AM
- The new block has editor styling and leaves the front end styling to the child theme
Screen Shot 2021-11-05 at 11 40 13 AM
- Newsletter is customizable with default fallbacks
  - The headline can be changed
  - The text description paragraph can be changed
  - The privacy policy legal text can be changed with the placeholder being replaced with themes' privacy policy link if set by replacing the placeholder `%%Privacy policy%%` (if no privacy policy is set, the placeholder is replaced with the text `Privacy Policy`
  - A background image can be selected or uploaded via the WP media library
Screen Shot 2021-11-05 at 11 38 33 AM
  - The mailchimp audience ID can be changed to target other TC mailchimp signups
Screen Shot 2021-11-05 at 11 39 11 AM
  - The ` Success Response` message shown in the form when a user successfully signs up can be customized; defaults to `Thanks and stay spooky!`.
  - The block width can be changed.
  - The block script *should* be conditionally added to each page depending on presence of the newsletter block in the widget area (if present).

### Additional Notes
- Class names are slightly different than in the CC newsletter update that was done in place o CC, css rules will need re-targeting there.
- A widget area needs to be added to the child themes in order to enable use of the newsletter block in desired places on the CC homepage and Single Posts pages

### Testing
- [x] Locally
- [x] Staging

### To Test
- [x] Can the newsletter block display in the editor with default content to match the text content in the CC newsletter design update?
- [x] Can these values be updated?
- [x] Can a bg image be selected and added?
- [x] Does the updated info save to the front end?
- [x] Does the form work (should redirect to a mailchimp page when signing up from a local, parent theme instance)?
- [x] Does the script only appear on pages that explicitly have the widget area with the newsletter block in it?


douglas-johnson:
> Does the form work (should redirect to a mailchimp page when signing up from a local, parent theme instance)?

Does this mean you were planning to leave the JavaScript to submit the form via AJAX in the child theme?

I assumed the block and the front-end script would stay together like the table of contents.

But, leaving them separate would help solve some the site specific customization like the response text when you sign up.

richdacuban:
> Does this mean you were planning to leave the JavaScript to submit the form via AJAX in the child theme?
- Yes. I was/am.
> I assumed the block and the front-end script would stay together like the table of contents.
> 
> But, leaving them separate would help solve some the site specific customization like the response text when you sign up.
- Thinking it over, I could see how this could be a problem, but the front end script would be difficult to make customizable via the block editor. I do see how I could build the mailchimp script into a block script.js file, I could move the already existing logic over.

douglas-johnson:
Regarding: "whether or not a script added in the block metadata has access to the block attributes"

No, it does not, but we control the markup, and the JS has access to that.

So either in the save component, or in the render_block filter you could add a data-* attribute to the form like `data-mailchimp-success=""` and our front-end JS could read it.

The other messages can be standardized. The only on that is custom for creepy catalog is success.

douglas-johnson:
There are not causing any problems currently, but I am concerned about leaving a real input and submit button inside the editor. We also don't need the `<label>` in this context.

Could they be mocked with divs instead? or at least make them more inert by adding `disabled` and removing all other form specific attributes.

Screen Shot 2021-11-05 at 12 14 52 PM

douglas-johnson:
@richdacuban Let me know when you'd like me to review again.

richdacuban:
@douglas-johnson Please take another look?
- in the editor, input is read-only, button is disabled, label is removed
- mailchimp script now lives in the blocks script.js and in the mailchimp js class which were both copied over from CC and re-targeted as needed
- script is dequeued for all pages and then re-enqueued if page `is_singular()` and `has_block( 'cata/newsletter-form' )` in the footer
- the form success response is customizable in the block sidebar
Screen Shot 2021-11-05 at 6 26 01 PM
Screen Shot 2021-11-05 at 6 25 49 PM

build/blocks/newsletter-form/src/script.js--output-path.js
```diff
@@ -0,0 +1 @@
...e-newsletter__description"},e.attributes.description),Object(c.createElement)("form",{action:""+e.attributes.mailchimpUrlBase+e.attributes.mailchimpUserId+e.attributes.mailchimpUrlAmp+e.attributes.mailchimpAudienceId,method:"post",name:"mc-embedded-subscribe-form",target:"_blank",validate:!0,"data-mailchimp":!0},Object(c.createElement)("div",{class:"cata-block-streaming-guide-newsletter__response",role:"alert"}),Object(c.createElement)("fieldset",{class:"cata-block-streaming-guide-newsletter__fieldset"},Object(c.createElement)("label",{for:"mce-EMAIL",class:"screen-reader-text"},"Enter Your Email"),Object(c.createElement)("input",{type:"email",value:"",name:"EMAIL",id:"mce-EMAIL",placeholder:"Enter Your Email",required:!0}),Object(c.createElement)("button",{class:"button is-primary is-filled",type:"submit"},"Subscribe"),Object(c.createElement)("div",{style:"position: absolute; left: -5000px;","aria-hidden":"true"},Object(c.createElement)("input",{type:"text",name:"b_"+e.attributes.mailchimpUserId+"_"+e.attributes.mailchimpAudienceId,tabindex:"-1",value:""})))),Object(c.createElement)("p",{class:"cata-block-streaming-guide-newsletter__details"},e.attributes.legalText))))))}})}])}});
```
douglas-johnson, line 1: I think this file is here by mistake. It has `--output-path.js` as the end of the filename.

build/blocks/newsletter-form/src/script.js--output-path.asset.php
```diff
@@ -0,0 +1 @@
+<?php return array('dependencies' => array(), 'version' => 'd8e343f74f68e3e716261c759765b75f');
```
douglas-johnson, line 1: Same with this one, seems like it shouldn't have `--output-path.asset.php` at the end of the filename.

blocks/newsletter-form/block.json
```diff
@@ -0,0 +1,72 @@
+{
+	"apiVersion": 2,
+	"name": "cata/newsletter-form",
```
douglas-johnson, line 3: I think `newsletter-form` is longer than we need. Just newsletter is enough. That should make the autogenerated class name shorter too.

blocks/newsletter-form/src/save.js
```diff
@@ -0,0 +1,90 @@
...
+export default function save(props) {
+
+	const blockProps = useBlockProps.save();
+
+	const { attributes } = props;
+
+	return (
+		<div { ...blockProps }>
+			<div 
+				class="cata-block-streaming-guide-newsletter"
```
douglas-johnson, line 34: I feel like the class names in both the Edit and Save components should be updated to match whatever WP produces based on the block name `cata/newsletter`. I think the base would be `wp-block-cata-newsletter`. Maybe there is a way to set that as a variable in here so other classes and be something like ``${base}__bg-img``

douglas-johnson (changes requested):
Looking good, I think there's some clean up, but seems on goal.

blocks/newsletter-form/block.json
richdacuban: updated all instances of `newsletter-form` to `newsletter`

build/blocks/newsletter-form/src/script.js--output-path.asset.php
richdacuban: removed that entire unintended build folder. I had a typo in the package.json; I forgot a space before the `--output-path...`

blocks/newsletter-form/src/save.js
richdacuban: I added a new attribute to the `block.json` file called `classnameBase` which contains the string `wp-block-cata-newsletter`. This attribute is used to build all the classNames in the edit and save components. Additionally, in both style rules files, I added a sass variable called `$base` equal to the same string and used that to build the class names in the style rules as well. Unfortunately, I still don't know how we can get the block attributes into the style rules in this framework so I defined the same sass var in both style rule files.

blocks/newsletter/block.json
```diff
@@ -0,0 +1,76 @@
...
+			"default": "Unsubscribe at any time. By subscribing, you agree to the terms of our %%Privacy Policy%%"
+		},
+		"successMessage": {
+			"type": "string",
+			"default": "Thanks and stay spooky!"
+		}
+	},
+	"editorScript": "file:./build/index.js",
+	"editorStyle": "file:./build/index.css",
+	"style": "file:./build/style-index.css",
```
douglas-johnson, line 74: Do we need this separate style sheet? It is output on the front-end as an inline style in the head. On it's own that's not bad, just not sure we want a lot of that. 

It only seems to have styles in it for the response. Does that need to be distinct?

richdacuban: > It only seems to have styles in it for the response. Does that need to be distinct?

Not necessarily, I simply wanted to ensure that the response wasn't visible on the front end. I can get rid of it.

richdacuban: Ok I removed it. It really wasn't necessary.

douglas-johnson:
@richdacuban after addressing other feedback, could you please do a tagged beta version and use that to test the integration with widgets on creepy catalog staging site? that would cover the checkbox for testing on staging.

blocks/newsletter/newsletter.php
```diff
@@ -83,24 +83,31 @@ function remove_editor_newsletter_script() {
...
 	}
-	if ( has_block( 'cata/newsletter' ) ) {
-		wp_enqueue_script( 'cata/newsletter', './build/script.js', array(), wp_get_theme()->get( 'Version' ), true );
+	if ( ! has_block( 'cata/newsletter' ) ) {
+		return;
 	}
-}
-
 
+	wp_enqueue_script( 'cata/newsletter', './build/script.js', array(), wp_get_theme()->get( 'Version' ), true );
```
douglas-johnson, line 110: @richdacuban you can just `wp_enqueue_script( 'cata-newsletter-script' );` here

douglas-johnson (changes requested):

richdacuban:
@douglas-johnson My final concern is that the `cata-newsletter-script` still appears on pages without the widget area or the newsletter block inside of it in the query manager script listing. Here is a screenshot of the About page running locally:
Screen Shot 2021-11-10 at 11 08 40 AM

I've confirmed that the dequeuing function does occur and before the functions that re-enqueue run. 

Testing further now.

douglas-johnson:
OK, `wp_enqueue_scripts` might be the wrong action. Maybe we need an action that is specific to block assets.

richdacuban:
> OK, `wp_enqueue_scripts` might be the wrong action. Maybe we need an action that is specific to block assets.

@douglas-johnson  Maybe something like [is_active_widget](https://developer.wordpress.org/reference/functions/is_active_widget/) could work, not sure yet if it applies to "widget areas" or not?

douglas-johnson:
@richdacuban Which of these is the case?

1. Script is not being removed at all by our action on `wp_enqueue_scripts`
2. Script is bring removed, but its being re-added in places it doesn't belong by our filters / actions on  `wp_body_open` and `widget_block_content`

richdacuban:
@douglas-johnson The fix you suggested worked, I applied it on the [CC PR #233](https://github.com/thoughtis/wpe-creepycatalog/pull/233). This plugin hasn't changed, the latest version is live on [creepy staging](https://creepystaging.wpengine.com)

douglas-johnson (approved):

PR #11, opened 2021-11-17
Issue#10/survive database sanitization

### LINKED ISSUE #10: Prevent form from being sanitized away by wp_kses
Once we got the newsletter block into production, we were unable to save the `<form>` to the database. So, let's just not save it.

I bet we can do it on render using render_block filter or by adding a render_callback to the block.

douglas-johnson:
### Related Issues

- Resolves #10 

### What Was Accomplished

- Updated Newsletter Block to include a `render_callback` argument so the output is server rendered rather than stored in the database.
- Removed the `save` component from the block script because there's no markup to save anymore. The attributes are still saved.
- Added the `Cata\Blocks\Newsletter\Renderer` class to actually do the server side rendering.

### How It Was Tested

- [x] Local ( dev site )
- [x] Local ( site using child theme )
- [x] Staging ( site using child theme )
- [x] Stakholder sign off

### How To Test

- [x] In the local dev site, the block can still be added and edited.
  - After I initially deleted the save component I had to `attempt block recovery` in the post editor and the "inactive widgets" section of the widget editor.

### Screenshots

From a site using a child theme:

Screen Shot 2021-11-17 at 3 22 53 PM
Screen Shot 2021-11-17 at 3 22 46 PM


richdacuban (approved):
Testing locally worked great!
Screen Shot 2021-11-18 at 10 30 29 AM
Screen Shot 2021-11-18 at 10 31 07 AM

PR #16, opened 2022-02-03
Remove Custom Webpack Configs

### LINKED ISSUE #15: Remove custom webpack configs since wp-polyfill is no longer included by default
According to this:
https://github.com/thoughtis/cata-blocks/blob/0.3.8/blocks/newsletter/webpack.config.js#L12

We were waiting for this to hit a tagged release of wp scripts because it remove wp-polyfill from front-end scripts without the need for a custom webpack config:
https://github.com/WordPress/gutenberg/pull/35436

### Acceptance Criteria
- [x] Blocks still build
- [x] wp-polyfill is not enqueued on the front-end as a requirment of our blocks.

douglas-johnson:
### Related Issues
- Resolves #15 

### What Was Accomplished
- Removed custom webpack configuations from Newsletter and Table of Contents blocks.
  - Those configs existed to prevent the `wp-polyfill` from being included as a dependency. Now that is the default behavior.
- Updated the package.json to remove reference to those configurations.
- Built both blocks using the updated configuration.
  - Only newsletter changed, even then it was only the version number. The distributable block files are identical, there's just fewer config files. 

### A note on @wordpres/scripts

I had to update my global install of @wordpres/scripts to get this to work. One of their dependencies was having a problem with a failing post-install script.

To get it working, I had to:
- Update to the latest version of Node through NVM
- run `export NODE_OPTIONS=--openssl-legacy-provider` to support Node 17.
- Install @wordpress/scripts globally with the `--ignore-scripts` setting


richdacuban (approved):
@douglas-johnson Looks good to me, after I got npm udated to 17 and wp-scripts updated to latest I was able to build to run the two npm commands to build the newsletter and table of contents blocks.

Checking them out in Cata via Local, the newsletter block needed to be recovered but it was successful. The table of contents block had no recovery needed and was fine.
Screen Shot 2022-02-03 at 4 34 28 PM
Screen Shot 2022-02-03 at 4 46 10 PM
Screen Shot 2022-02-03 at 4 46 27 PM
Screen Shot 2022-02-03 at 4 45 52 PM
Screen Shot 2022-02-03 at 4 47 56 PM
Screen Shot 2022-02-03 at 4 53 08 PM

PR #19, opened 2022-03-15
Update blocks for compatibility with @wordpress/scripts version 22.x

douglas-johnson:
### Related Issues

I started working on adding a block and found that the structure produced by `npx @wordpress/create-block` was different than our existing blocks. Investigating showed that the template had been changed so that `block.json` is included in the `/src` folder of each block and later copied to `/build`. I thought new and existing blocks should match so there's no confusion about how to build them.

### What Was Accomplished
- Updated to latest version of `@wordpress/scripts`
- Updated all blocks so the block.json file is in the `/src` folder.
- Updated all block php files to load the block from the `/build` folder.
- In the current version of  `@wordpress/scripts` builds are only handled correctly if they are called from the specific block's folder and the block has a package.json file.
  - Updated all blocks so they have a package.json file that includes their build script.
  - Removed custom build scripts from the top-level package.json.
- Updated readme to reflect these changes.
- Rebuilt and tested all blocks after updates and reorganization.
- Added `.nvmrc` set to an `lts` version of Node. 17+ was having issues.


richdacuban (approved):
Was able to run `npm run build` in each folder (after deleting `/build`) and saw no errors!

PR #21, opened 2022-03-31
Shop

### LINKED ISSUE #20: Questions about shop block
# Step 1 (current)
- How to turn Creepy Catalog merch into a block?
  - [x] Create a "woocommerce products from another site" block.
  - [x] Get products from WP API.
  - [x] How to store something, and keep it up to date?
    - API fetch response stored as url-hashed transient, transient updated site-wide by single block editor api fetch
  - [x] How to do that without both client-side and server-side rendering?
    - Block uses dynamic `render_callback` which renders url-hashed transient product data on the front end; if no transient it does not display
  - [x] How to do that if there is more than one in a block?
    - Multiple blocks with different queries and therefore different stored url-hashed transient data can be displayed side by side on the same post

# Step 2 ( future )
- How to do that if the block is in a widget or a FSE template instead? Basically how to avoid relying on context?
- How to bust cache or avoid wasting resources when a block is eliminated?

richdacuban:
### Related Issue(s)
- Closes #20 

### Accomplishes
- Adds an external products block in the editor which queries the ShopCatalog API to retrieve a customizable number of products from a specific product category.
  - The query stores the API response as a `url-hashed` transient.
  - Uses abortController to enable canceling of request if needed.
  - Uses dynamic render callback to render the transient product data on the front end.
  - Checks is user is logged in and has `edit_posts` capabilities before executing external fetch.
  - Provides basic styling in editor and front end for child theme to build upon.
  - Does not display any front end output if no products found/product array empty.
- To update the data, an editor simply needs to re-fetch the data using the same category and quantity as before and immediately the `url-hashed` transient for that query will be updated site-wide.

### Tested
- [x] locally
- [ ] staging

### Screenshots
Screen Shot 2022-03-31 at 4 25 33 PM
Screen Shot 2022-03-31 at 4 25 48 PM


blocks/shop/shop.php
```diff
@@ -0,0 +1,232 @@
...
+}
+
+/**
+ * Render Product
+ * 
+ * @param stdClass $product
+ * @return string
+ */
+function render_product( stdClass $product ) : string {
+	$link   = esc_url( apply_filters( 'creepycatalog_product_block_link', $product->permalink ) );
```
douglas-johnson, line 89: This filter name should probably change to cata_ from creepycatalog_ and use whatever block name we decide on.

cata-blocks.php
```diff
@@ -85,3 +85,33 @@ function cata_blocks_proxy_request( WP_REST_Request $request ) {
...
 
 }
+
+/**
+ * Add UTMs to Product Block links
+ *
+ * @param string $link Link to be modified.
+ * @return string $new_link Modified product link.
+ */
+function cata_add_product_link_utm( string $link ) {
```
douglas-johnson, line 95: Sorry I should have been more specific. I think this would be left on the site that's actually using the blocks. In the plugin, we're just making it available to use.

blocks/shop/shop.php
richdacuban: Ok, so I added the `xxx_add_product_link_utm` filter and callback to the shop-block from CC into the .cata-blocks.php` file (for now, at least). The callback and filter names are `cata_` prefixed as well now: `cata_product_block_link`, `cata_add_product_link_utm`.

I only needed to modify the part of the code which adds the query arg for `utm_campaign` to the link url. I simply added this line and set the `utm_campaign` query arg equal to the value of this:
```php
$site_name = strtolower( preg_replace("/\s+/", "", get_bloginfo( 'show' ) ) );
```
So, when testing this on CC locally, before adding the `strtolower()` and the `preg_replace()`, the return value of `get_bloginfo( 'show' )` was `Creepy Catalog` which ended up in the URL just like that. After, the output in the product link was `creepycatalog`, as expected.

cata-blocks.php
richdacuban: Oh, ok. I see.

So the callback function should be in the child theme, the same as before, but we rename it `cata_add_prod...`  using the cata prefix?

So the [line in render_product() in the shop block that applies the filter](https://github.com/thoughtis/cata-blocks/blob/eb49536c3407d474cb5f73f7cdcc1018e2c0aeff/blocks/shop/shop.php#L89) can call it from the name `cata_product_block_link`?

douglas-johnson: Right, exactly

richdacuban: Ok, so this will require a small PR and update to CC (and any other child theme) before this new block is usable then, right?

douglas-johnson: Yes

cata-blocks.php
```diff
@@ -27,5 +31,57 @@
 require_once __DIR__ . '/blocks/kicker/kicker.php';
 require_once __DIR__ . '/blocks/newsletter/includes/class-renderer.php';
 require_once __DIR__ . '/blocks/newsletter/newsletter.php';
+require_once __DIR__ . '/blocks/shop/shop.php';
 require_once __DIR__ . '/blocks/table-of-contents/table-of-contents.php';
 require_once __DIR__ . '/patterns/trivia/trivia.php';
+
+add_action( 'rest_api_init', function () {
```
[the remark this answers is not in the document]
richdacuban: Done!

cata-blocks.php
richdacuban: Ok, done. I have the code ready to be submitted as a a PR on the CC from my local testing (once the cata-blocks plugin updated is merged to main)

richdacuban: @douglas-johnson I think this block is ready to be merged (and then vers bumped).

Do you concur?

blocks/products/products.php
```diff
@@ -0,0 +1,231 @@
...
+}
+add_action( 'init', __NAMESPACE__ . '\\create_block_shop_block_init' );
+
+/**
+ * Products Block Render Callback
+ * 
+ * @param array $block_attributes
+ * @return string
+ */
+function render_callback( $block_attributes) {
```
douglas-johnson, line 40: I feel this is isn't sufficiently unique. We could easily have another `Cata\Blocks\render_callback` for another block.

includes/class-products-proxy.php
```diff
@@ -0,0 +1,75 @@
...
+		$status = wp_remote_retrieve_response_code($response);
+
+		if ( 200 !== $status ) {
+			return new WP_Error( 'proxy_error', 'Error making remote request', array( 'status' => 404 ) );
+		}
+
+		$body = wp_remote_retrieve_body( $response );
+
+		// No expiration time
+		set_transient( 'cata-blocks-shop-proxy-req-result-' . $hash, $body );
```
douglas-johnson, line 71: Have we checked how long the `$hash` is?

WP says transient keys are max 172 characters and the prefix here is 34 characters.

https://developer.wordpress.org/reference/functions/set_transient/

blocks/products/products.php
richdacuban: How about `products_block_render_callback`

PR #45, opened 2022-10-17
Issue#42/remove product block site specifics

### LINKED ISSUE #42: Products Block: remove font-family specifics
The products block has font styles that are intended for a specific site, but they should be more generic and the site using the plugin should update their theme to apply font styles

- This font family can just be sans-serif system fonts:
  - https://github.com/thoughtis/cata-blocks/blob/0.6.3/blocks/products/src/style.scss#L60

### LINKED ISSUE #43: Products Block: Allow editor to omit the price
### Background
In previous contexts where all products had the same byline, we wanted to omit the byline. Now we're interested in omitting the price in some contexts.

### Details
- Add `display_byline` attribute, defaulting to true
- Use that attribute in the JSX for the editor
- and in the PHP renderer for the front-end

### References
- https://github.com/thoughtis/cata-blocks/blob/0.6.3/blocks/products/src/block.json#L43-L46
- https://github.com/thoughtis/cata-blocks/blob/0.6.3/blocks/products/src/edit.js#L134
- https://github.com/thoughtis/cata-blocks/blob/0.6.3/blocks/products/includes/renderer.php#L45

### LINKED ISSUE #44: Products Block: Image aspect ratio in widget editor
### Summary

The images in the products block look vertically stretched and horizontally condensed in the widget editor.

### Details
- The widget editor seems to be missing some things from our common reset, which makes sense.
  - `<figure>` has margin which is reducing available inline size
  - `` is assigned an inline-size by our CSS but block-size is not set to auto by default.

### Screenshots
Screen Shot 2022-10-13 at 9 51 04 AM

douglas-johnson:
### Related Issues
- Resolves #42 
- Resolves #44 
- Resolves #43

### What Was Accomplished
- Added `display_price` attribute
- Updated the product block's edit component to control the `display_price` attribute with a toggle and show/hide the price depending on its value.
- Updated the product block's render callback to show/hide the price depending on the value of `display_price`
- Added separate components from Byline and Price to simplify the edit component
- Removed multiple styles related to font families and size that should be determined by a child theme.
- Added styles to `editor.scss` to ensure figures and images fill the available space

### How It Was Tested
- [x] Local environment for parent theme
- [x] Local environment for child themes

### How To Test
- [x] Show/hide price works in both the editor and rendered post content
- [x] When using the widget editor, image are no longer stretched vertically. See the before picture in the linked issues.


jessica-townsend (approved):
Approved!

PR #51, opened 2022-11-14
Refactor products block for more small components

### LINKED ISSUE #47: Products Block: Refactor for more small components, prepare for updates to product selection
### Background
The current version of the products block contains a lot of firsts for us. It was our first attempt to use API data in a block and keep it up to date, it was also our first time using apiFetch in the editor. We learned a lot when making it, and even more when we did it a second time on another project. We have integrated the things we learned about the server-side aspect of the block ( keeping the API data up to date ) but haven't refactored the `edit.js` file much.

### Details

- The edit component should be broken down into smaller components and named functions
- The URL composing process should be its own function or component
- Layout can be turned into its own component
- We can use middleware on apiFetch to do proxying and request caching
- The coupling with Creepy Catalog styling can be further reduced.
- We might not need the AbortController, or if we keep it, it might be possible to simplify and made easier to test.

douglas-johnson:
### Related Issues
- Resolves #47 

### What Was Accomplished
- Added middleware to `apiFetch` for caching
  - This caching is in JavaScript memory. It helps ensure a quick reponse if the user is flipping through options and hits a previously selected combination.
  - This middleware is opt-in using the `{ cache: true }` option.
  - `cache` is a real `fetch` option and true is not a valid value. Our cache  option for`apiFetch` is unset before proceeding to `fetch`
- Added middleware to `apiFetch` for proxying
  - We were already doing this proxying, but it was just done by modifying the request URL. 
  - This version using middleware is opt-in using the `{ proxy: true }` option.
  - It proxies the WP REST API request through the server running the editor in order to resolve permissions issues. Only public data is available in this way.
- Added `Products` component to remove large component loop from edit component. If we want better loading or error states, we can re-apply them to the Products component, or wrap it with a HOC rather than keeping it all together in the edit component.
- Added `Product` component to simplify `Products` component.
- Removed the `shopUrl` variable that created an extra step before setting the `query_url`.
- Removed the Fetch and Cancel buttons. As a result of using `query_url` directly, the request is made on render.
- Updated the request and response pattern to use the `query_url` attribute directly rather than using state to avoid updating the attribute until we have a successful response.
- Used more hooks to create a more obvious chain of events from updating the query options -> make a request -> updating products.

### How It Was Tested
- [x] Local plugin environment
- [x] Local site environment
- [x] Remote site environment

### How To Test
- [x] Ensure this is backward compatible by using the plugin development environment to create a product block while using the version currently on the main branch. Then switch to this branch and reload the editor. If its working correctly, everything is normal.


douglas-johnson:
@jessica-townsend I finished my own testing of this. Should be ready for review.

jessica-townsend (approved):
Approved, product block did not change when switching branches and all options worked as expected.

PR #54, opened 2022-11-15
Issue#50/add new product block style

### LINKED ISSUE #50: Products Block: Layout Options
### Background

We are interested in more layout options for the products block. Currently we have a single grid-like layout.

Designs are not finalized, but we're looking into a "reel" type layout that displays a single row of products and allows scrolling when the content of that row is wider than its container.

### Questions

How do we want to apply this layout change?

https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/block-controls-toolbar-and-sidebar/

#### Thought Catalog example

The Thought Catalog REST API block has two layout options. We apply them using a select input in the sidebar panel.

All styles are handled by us and the layouts are coded twice, in React and PHP, and the resulting HTML is saved to the post.

https://github.com/WordPress/gutenberg/blob/HEAD/packages/block-editor/src/components/inspector-controls/README.md

Screen Shot 2022-11-10 at 1 47 16 PM

#### WordPress core example

The core query loop block uses the formatting toolbar for basic layout options like grid and list.

Further options are handled through the post template block. I do not totally understand how it works. It saves the query parameters and the options for each child of the post template block. So, it's possible to utilize things like font-size controls.

https://developer.wordpress.org/block-editor/reference-guides/slotfills/plugin-block-settings-menu-item/

https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/query
https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/query/edit/query-toolbar.js
https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/post-template

Screen Shot 2022-11-10 at 1 49 52 PM

jessica-townsend:
### Relevant Issues
- Closes #50 

### What Was Accomplished
- Added new style option called "Reel" to the products block
- Added CSS to support the "Reel" style option

### How It Was Tested
- [x] Locally on Cata theme
- [x] Locally on Collective World
- [x] Develop on Collective World


douglas-johnson (changes requested):
This has some changes in it from 0.6.4 and 0.6.5, can you rebase from main to get it in sync? Let me know if I can help.

douglas-johnson (approved):

Issue #52, opened 2022-11-14
Product Block: Grid style


douglas-johnson:
### Background

While thinking through #50 we realized that the grid styling that is applied by default in the editor would need to be undone to allow the reel style to work correctly. Can we make a grid style that is the default, and keep the layout styles separated that way?

### Details

- Add a `grid` style to the products block which is the default style
- Move the existing styles into the resulting `is-style-grid` CSS class

jessica-townsend:
@douglas-johnson something to note here...

After adding support for the new reel style, two style options appear in the editor:

<img width="279" alt="Screen Shot 2022-11-11 at 10 43 36 AM" src="https://user-images.githubusercontent.com/70295707/201720063-fd6c577a-83b4-473b-8cd6-1605d1cdebe7.png">

If the "Default" style is selected, the class `is-style-default` gets added to the products wrapper on the frontend.

douglas-johnson:
@jessica-townsend I have a branch together here: https://github.com/thoughtis/cata-blocks/compare/issue%2352/grid-style

Once we merge your work on #53 I'll rebase it.

It currently uses `:not(.is-style-reel)` because that will work whether we merge this first or the reel style first.

I can refactor it to use `.is-style-default` after the reel style exists.

PR #55, opened 2022-11-15
Issue#52/grid style

douglas-johnson:
### Related Issues
- #50
- #52 

### What Was Accomplished
- Moved the default layout of the products block ( a grid layout ) into `.is-style-default`
- Updated the default WC API query parameters to include `status=publish` after we noticed sorting by menu_order in ascending order can include private products, resulting in an empty API response.


jessica-townsend (approved):
Approved!

Issue #46, opened 2022-10-31
How to provide more options for product selection?


douglas-johnson:
In several cases we want more control over which products are displayed by the products block. The following feature requests have come up:

- Products in a specific author taxonomy, or any taxonomy other than product category really.
- Ability to exclude pre-orders.
- Manually selected products.

PR #58, opened 2022-11-16
Add tag filter to product selection

douglas-johnson:
### Related Issues
- #46 
- Long term I would like to make something where clicking the `[+]` button lets you add multiple filters like category, author and tags. I think the tipping point for that will be authors. For now its just one more text input for product tags.

### What Was Accomplished
- Updated the products block to add a `tag` attribute that can contain a product_tag taxonomy term id, or multiple comma separated terms.
- Updated the API query URL build to optionally add either category or tag parameters to the URL.
- Removed the default value of the category attribute since it is now optional.

### How It Was Tested
- [x] Local plugin dev site
- [x] Local site with child theme
- [x] Remote site with child theme

### How To Test
- [x] Filtering by product tag produces expected results.
- [x] Filtering by product category still produces expected results.
- [x] Filtering by both a product tag and category uses an `AND` type query where products must be assigned to both terms.


jessica-townsend (approved):
I tested locally by setting categories & tags by themselves, and then setting both a category and a tag. Everything worked for me as expected.

PR #65, opened 2022-12-08
Add coauthor taxonomy filter, use methods to prevent API requests for every character entered

### LINKED ISSUE #48: Filter by author taxonomy
### Background

We would like to be able to filter products in the products block by the author / creator.

We use CoAuthors Plus on Shop Catalog, so on some level it should be possible to use the `author` taxonomy in the REST API.

By default, CAP does not fully support the REST API, but we have created more fully featured support through [Cata CoAuthors Plus](https://github.com/thoughtis/cata-co-authors-plus/). This work may require updates to that plugin as well.

#### API Support

Here you can see an example of the varying levels of REST API support. Shop Catalog responds with a forbidden error, while Thought Catalog shows some information about the taxonomy because it's running our plugin.

https://shopcatalog.com/wp-json/wp/v2/taxonomies/author
https://thoughtcatalog.com/wp-json/wp/v2/taxonomies/author

#### Ids

Even with API support, an `author` taxonomy term_id as provided by CoAuthors Plus is somewhat difficult to obtain. We typically deal with the slug as a unique identifier but the WP REST API only accepts term_ids when attempting to list a post type by a taxonomy term.

We might need to expand the read / view access for the author taxonomy on Shop Catalog so that unauthenticated users working in the editor of another site are to list author terms through the API.

douglas-johnson:
### Related Issues
- Resolves #48 
- https://github.com/wpcomvip/shopcatalog/pull/784

### What Was Accomplished
- Added an attribute to the products block for `coauthor`
  - Based on code that is awaiting review on Shop Catalog, it should be possible to use this value to filter WooCommerce products by the `author` taxonomy.

#### Efforts to prevent API requests on each input change
- Updated products block edit component to include a clean up function in `fetchData`.
  - This aborts a fetch request if the relevant state changes before the request finishes.
- Added a `DebouncedTextControl` component
  - This component waits to provide updates to its `onDebouncedChange` function until the provided timeout has expired. The timeout clock starts again any time the relevant input changes.

### How It Was Tested
- [x] Local plugin environment
- [x] Local theme environment
- [x] Remote theme environment

### How To Test
- [x] Entering an author's URL slug correctly filters the returned products.
- [x] Entering multiple author URL slugs separated by a comma uses an `OR` type filter where the results can be from either author.
- [x] Entering an author URL slug and a category id uses an `AND` type filter where the results must be in both taxonomies.


blocks/products/build/block.json
```diff
@@ -14,12 +14,16 @@
 	"attributes": {
 		"shopCatalogApiUrlBase": {
 			"type": "string",
-			"default": "https://shopcatalog.com/wp-json/wc/v3/products/"
+			"default": "https://shopcatalog-com-develop.go-vip.net/wp-json/wc/v3/products/"
 		},
```
jessica-townsend, line 18: @douglas-johnson Should this be the live shop catalog URL instead of the development site?

douglas-johnson: Yes, thanks for reminding me. I thought I discarded that.

douglas-johnson: Oh, I did discard it, but I forgot to build it. Just pushed.

jessica-townsend (approved):

Issue #12, opened 2022-01-26
Table of Contents and Heading Ids


douglas-johnson:
As of WordPress 5.9, headings are auto-assigned an id.

We may not need to do that any more in the front-end script for the table of contents.

Or we might still need to ensure the id is there because not all content will be recently edited.

douglas-johnson:
I think we could mark the auto id generation as deprecated, and remove it in a future version.

PR #66, opened 2022-12-15
Enable "generateAnchors" setting in block editor when table of contents block is supported

douglas-johnson:
### Relate Issues
- #12 

### What Was Accomplished
- Updated the registration of the Table of Contents block to add a filter on `block_editor_settings_all`
- Added function `enabled_generate_anchors` to set `generateAnchors=true` in the block editor settings.

### How It Was Tested
- [x] Local plugin dev site
- [x] Local theme dev site

### How To Test
- [x] Ids are generated for headings, as shown in the screenshots below
- [x] Once an id is generated, you can edit the id, and further edits to the heading text will not change the id

### Open Questions
- This doesn't fully solve the issue because we're not abe to turn off our script that handles ids. I think the remaining task would involve generating anchors all existing Tables of Contents.

#### Auto generated id
Screen Shot 2022-12-15 at 9 30 47 AM

#### Auto generated id that I edited

Screen Shot 2022-12-15 at 9 30 36 AM

### DIFF
--- a/blocks/table-of-contents/table-of-contents.php
+++ b/blocks/table-of-contents/table-of-contents.php
@@ -17,6 +17,7 @@ function register_toc_block() : void {
 		return;
 	}
 	register_block_type( __DIR__ . '/build' );
+	add_filter( 'block_editor_settings_all', __NAMESPACE__ . '\\enabled_generate_anchors' );
 }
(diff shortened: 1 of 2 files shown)


jessica-townsend (approved):
Looks good!

PR #71, opened 2022-12-19
Issue#68/table of contents summary

### LINKED ISSUE #68: Allow editor to change the summary for Table of Contents
### Background

The links in the current Table of Contents are generated on the front-end; becaue of that we also client-side render the HTML around the links so that you don't wind up with an empty `<details>` and `<nav>` taking up space.

The `<summary>` element in `<details>` literally says "Table of Contents", but we have designs for a new site where it says "Content:" instead.

### Details
- How can we change the summary in the editor, and still conditionally render based on whether a TOC was produced?

douglas-johnson:
### Related Issues
- Resolves #68 

### What Was Accomplished
- Updated the Table of Contents block to use a `RichText` component for control of the `<summary>` content, plus the myriad settings you need in order to make hitting `Enter / Return` start a new block after this one.

### How It Was Tested
- [x] Local plugin dev site
- [x] Local theme dev site

#### Existing Content

Start with the `main` branch

- [x] Create a post with a table of contents and some `h2` headings.

Switch to this version, available at the tag `0.7.6-beta1`

- [x] Without modifying the post content, view the post. The new client-side script should still work to produce the TOC.
- [x] Reload the editor. You should see an "Attempt Block Recovery" button. Clicking it should work, and "Table of Contents" should be correctly inserted as the default.

#### New Content

- [x] When inserting a TOC block, "Table of Contents" is used as the default summary.
- [x] Modifying the summary is saved, and is reflected when viewing the post.
- [x] The modified summary persists in the editor.

### Open Questions
- @jessica-townsend In the editor, how should we communicate that the TOC is client-side rendered? Should we placehold the list of links with something? Right now it just has the summary.


jessica-townsend:
> In the editor, how should we communicate that the TOC is client-side rendered? Should we placehold the list of links with something? Right now it just has the summary.

I think something that says "Links will be rendered here" below the summary would work best

douglas-johnson:
> I think something that says "Links will be rendered here" below the summary would work best

@jessica-townsend How about this?

Screen Shot 2022-12-20 at 10 20 19 AM

jessica-townsend:
That looks great!

jessica-townsend (approved):

PR #74, opened 2023-02-21
Add product_cat_exclude parameter

### LINKED ISSUE #49: Exclude products by product category
### Background

We'd like to be able to exclude a product category or categories from REST API queries. Or first you cause is to make separate Book and Non-Book product blocks on author profiles.

The regular WordPress REST API has a `categories_exclude` parameter, but we're actually dealing in product categories which is not the same taxonomy.

https://developer.wordpress.org/rest-api/reference/posts/#list-posts

The WooCommerce API doesn't mention excluding product categories.

https://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-products

### Questions

- Is there a parameter that's part of the WordPress Core REST API that would let us exclude products from specific product categories?
  - Prove it by applying it to this endpoint: https://shopcatalog.com/wp-json/wc/v3/products

douglas-johnson:
### Related Issues
- Resolves #49 

### What Was Accomplished
- Added `product_cat_exclude` parameter as a way to add a `NOT IN` taxonomy query to WooCommerce REST API requests. This is supported by a custom plugin added to Shop Catalog.
- Updated the parameter names in `onChange` events using a `nextParamName` naming convention to avoid any potential for collision with existing state.

### How It Was Tested
- [x] Local plugin dev site
- [x] Remote site dev environment

### How To Test
- [x] Using the Product Author Slug and Exclude Product Id options together can display the products associated with an author other than their actual books.


jessica-townsend (approved):
Works great for me locally

PR #80, opened 2023-03-09
Issue#75/toc anchor text

### LINKED ISSUE #75: Adjust what text is included as a link in Table of Contents
### Details
We need a way to adjust the TOC block so that only part of the text is a link. Example here:
Screen Shot 2023-03-06 at 1 13 32 PM
https://www.allure.com/story/zodiac-sign-personality-traits-dates

How can we implement this?

jessica-townsend:
### Related Issues
- Closes #75
- Regex to be added to the child theme https://github.com/wpcomvip/thoughtnet/issues/348

### What Was Accomplished
- Updated TOC block script to move certain heading text outside of the anchor link element and display it after the link.

### How It Was Tested
- [x] Local plugin dev environment
- [x] Remote site dev environment

### How To Test
Use plugin version 0.7.9-beta1
- [x] Add a table of contents block in the editor.
- [x] Create a H2 heading in the editor like this: `Aries (January 1st - February 1st)`.
- [x] On the frontend, only the text before the open parentheses is a link.
- [x] Other links are unchanged.


blocks/table-of-contents/src/script.js
```diff
@@ -50,14 +51,17 @@ function handleBlock( block ) {
 		}
 	});
 
+
 	/**
 	 * Render
 	 * data -> links -> listItems -> list -> nav
 	 */
+	renderData = data
```
douglas-johnson, line 59: Can this either be assigned using `const` or put back inside `renderList` so it doesn't need to be assigned to a variable?

blocks/table-of-contents/src/script.js
```diff
@@ -169,10 +173,24 @@ function renderList( children ) {
...
-	listItem.appendChild( child );
-	return listItem;
+function renderListItem( regexAppend ) {
+	return (child) => {
+		const listItem = document.createElement( 'li' );
+		let appendText = '';
+		
+		if( isNonEmptyString( regexAppend ) ) {
+			appendText = (new RegExp( regexAppend, 'mi' )).exec( child.textContent );
+		}
```
douglas-johnson, line 183: Can this be a function like `getAppendText( regexAppend, child.textContent)` so it can be assigned to const instead of let?

douglas-johnson (changes requested):
This link has empty space at the end.

Screenshot 2023-03-09 at 2 21 29 PM

blocks/table-of-contents/src/script.js
```diff
@@ -233,6 +243,20 @@ function getRegexFunction( regexPattern ) {
...
+ *
+ * @param {String} regexAppend
+ * @param {String} textContent
+ * @return {String}
+ */
+function getAppendText( regexAppend, textContent ) {		
+	if( isNonEmptyString( regexAppend ) ) {
+		return (new RegExp( regexAppend, 'mi' )).exec( textContent );
+	}
+	return textContent
```
douglas-johnson, line 257: shouldn't this return an empty string if there's no regexAppend?

jessica-townsend: You're right, not sure what I was thinking here

douglas-johnson (approved):

PR #81, opened 2023-03-23
Issue#76/toc open

### LINKED ISSUE #76: Table of Contents always open
In our initial use case of the Table of Contents, we wanted it closed by default. That's why we chose a details element. It was not emphasized and keeping it closed while the links were rendered was good for core web vitals, see #28 

But now every implementation is open by default, and in some the ability to close it is moot.

### Questions
- Should we provide an option to keep it open?
- Or should open be the default and making it collaspable requires opt-in?
- If we keep it open, should the elements we use change? We don't need details if there's no interactivity.

douglas-johnson:
### Related Issues
- Resolved #76 

### What Was Accomplished
- Added three levels of control for the Open/Close Behavior of a Table of Contents
  - Always Open
  - Start Open
  - Start Closed
- The "Always Open" setting uses `div` elements instead of `details` and `summary`. Other options for keeping the TOC open like adding `onclick="return false;"` removed default functionality. This way should be more accessible since the elements behave as expected.

### How It Was Tested
- [x] Local plugin dev site
- [x] Local theme dev site
- [x] Remote theme dev site

### How To Test

Create a table of contents in a post using each of these earlier versions, then switch to this branch and ensure Attempt Block Recovery works.

- [x] A version tag 0.7.5 or earlier
- [x] Main branch


jessica-townsend (approved):
Looks good to me!

PR #83, opened 2023-04-13
Issue#82/toc emit event

### LINKED ISSUE #82: Emit custom event after TOC block is rendered
### Background
In https://github.com/thoughtis/cata/issues/148, we ran into an issue where the TOC block is rendered after the link-related JS is run, so any links in the TOC block don't have the correct styles.

### Details
- Emit a custom event after the TOC block is rendered so that the link script can detect it and re-run
- https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent

jessica-townsend:
### Related Issues
- Closes #82 

### What Was Accomplished
- Dispatches a custom event after the links for the TOC block are rendered

### How It Was Tested
- Locally


blocks/table-of-contents/src/script.js
```diff
@@ -1,7 +1,10 @@
 /**
  * Table of Contents
  */
-document.querySelectorAll( '.wp-block-cata-toc' ).forEach(handleBlock);
+const tocBlocksRendered = new CustomEvent( 'tocBlocksRendered' );
```
douglas-johnson, line 4: Do we use this variable for anything else? Could line 7 be condensed to?

`document.dispatchEvent( new CustomEvent( 'tocBlocksRendered' ) )`

douglas-johnson (approved):

PR #86, opened 2023-05-05
Issue#84/faq block

### LINKED ISSUE #84: FAQ Block
### Background

We are currently using this plugin for FAQs on Creepy Catalog. They are working and are accepted by Google, which is great, but there are a few things we'd like to improve in our own version before reusing it across out network.

https://wordpress.org/plugins/faq-block-for-gutenberg/
https://plugins.trac.wordpress.org/browser/faq-block-for-gutenberg/tags/2.4/includes/class-faq-block-for-gutenberg.php
https://plugins.trac.wordpress.org/browser/faq-block-for-gutenberg/tags/2.4/includes/class-faq-block-for-gutenberg.php#L116
https://developers.google.com/search/docs/appearance/structured-data/faqpage

This is a new block. I think our [How To Add A Block](https://github.com/thoughtis/cata-blocks#how-to-add-a-block) documentation in the readme is still valid.

#### Don't parse blocks if we don't have to
- I think we could use an action on `wp_body_open` to set up an FAQ queue.
- Then in an action on `render_block`, FAQs could be pushed into the queue.
- Lastly in `wp_footer` we could output the structured data.

#### Support paragraphs and lists in answers
- Google's support for FAQ structured data includes a limited number of HTML elements, including lists which we've wanted in FAQ answers before.
  - https://developers.google.com/search/docs/appearance/structured-data/faqpage
- In the current plugin version, the answer is a single RichText input, so lists would require writing the html manually.
- It should be possible for the answer portion of the FAQ block to instead utilize InnerBlocks to output multiple blocks, while limiting those avaialable to the ones supported by structured data.
- https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/nested-blocks-inner-blocks/
- https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/nested-blocks-inner-blocks/#using-a-react-hook

#### Question should be a heading block, with all the normal block settings
The question is always an h4 in the FAQ block plugin, that should be a real heading block where the heading level can change.

#### Only produce URLs if required and real
The FAQ block plugin produces URLs like `https:\/\/creepycatalog.com\/new-french-extremity-movies\/#644fceb9d2211` in the structured data. That hash does not exist as anchor in the posts. If those URLs are required by Google, let's make them real, otherwise they can be omitted.

### Other features
- Our block should have a `transform.from` in its settings to transform a existing FAQ block into one of our own.
  - https://developer.wordpress.org/block-editor/reference-guides/block-api/block-transforms/
  - This transformation can ignore any of the color settings, we just need the content.

### Open questions
- In the current version, the question is always displayed inline before the answer. That's how we're handling FAQs on other sites as well. But, it gets weird if the first element in the question is a list.
- We have other implementations of Question and Answer formats, for example trivia on Creepy Catalog, FAQs about a property on Clearing Farm and author interview questions on Collective World. Is there any reason to think this FAQ block that outputs structured data should be a wrapper around a Question and Answer block that can handle more use cases?

jessica-townsend:
### Relevant Issues
- Closes #84 

### What Was Accomplished
- Added new FAQ block with a heading for the question and lists/paragraphs for the answer.
- Injected an FAQ Page schema to the footer without using `parse_blocks`
- Added an Inline custom style for the FAQ block, which is also reflected in the editor
- Added ability to transform old FAQ blocks into new FAQ blocks

### How It Was Tested
- [x] Locally on parent theme
- [x] Staging on child theme

### How To Test
- [x] Custom FAQ block can be added through the editor (icon for our block is a question mark)
- [x] Only heading and list blocks are available to add in the FAQ block
- [x] Selecting the `inline` style for the FAQ block displays the question heading and the next paragraph in the block side by side in the editor. Lists immediately following the question heading are not displayed inline.
- [x] An old FAQ block can be transformed into our new FAQ block through the editor
- [x] Google's Rich Results Test detects a valid FAQ Page item
- [x] Answer text in the schema is stripped of any [unsupported tags](https://developers.google.com/search/docs/appearance/structured-data/faqpage#answer) and all attributes besides `href` for anchor tags

### Open Items
- Need to update the parent theme to properly display the inline FAQ block style 
- Need to update child themes to update question heading default font size


douglas-johnson:
I foresee an issue where people accidentally delete the heading, don't know how to get it back with the correct class name, and it silently doesn't output the schema.

In the latest version of Gutenberg, group gets its own `allowedBlocks` attribute. So when that hits WP core, we could use templateLock on the FAQ. The heading and group would be locked in place and the group could only contain the blocks we allow.

https://github.com/WordPress/gutenberg/releases/tag/v15.7.0
https://github.com/WordPress/gutenberg/pull/49128

So I'm suggesting:
- bring back the group, the way you originally had it
- update the description ( and possibly placeholders ) to advise that people only use paragraphs and lists in the answer
- plan to implement allowedBlocks when it hits WP core

douglas-johnson:
After my previous comment I remembered why we wanted to get rid of the group. I think we can have both template locking and the styling we want, it just might take multiple blocks. Let's discuss today.

jessica-townsend:
@douglas-johnson could we just lock the heading block individually and leave everything else unlocked?
https://developer.wordpress.org/block-editor/reference-guides/block-api/block-templates/#individual-block-locking

douglas-johnson:
@jessica-townsend yes, christ! if that works that's amazing. I saw someone doing that on github but I couldn't find it in the docs.

jessica-townsend:
Cool, I have that up on staging if you'd like to review it @douglas-johnson

douglas-johnson:
The locking works great. I'm glad it was so simple, thank you for figuring that out.

I would you like to discuss just one more thing. I think its unintuituve to get out of the FAQ block using the keyboard. I had to hit my up arrow so the FAQ block was selected, then hit enter.

If I just hit enter, it adds new paragraphs in the FAQ.

Screenshot 2023-05-09 at 9 32 32 AM

Take a look at how the group block works. If you hit enter in an empty paragraph, it moves you out of the group block.

I thought that was handled by an `onSplit` function but it turns out that might be from this `__experimentalOnEnter`

https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/group/block.json#L24

Can you try declaring support for it in the FAQ block.json and see if it works?

jessica-townsend:
@douglas-johnson this worked for me. I pushed to staging too

PR #91, opened 2023-05-10
Issue#90/fix broken schema text

### LINKED ISSUE #90: Fix broken question text for the FAQ block schema
### Background
We recently launched a new FAQ block that generates an FAQ page schema for rich results in Google. When testing the transform functionality on production, we noticed that the question text in the schema gets broken up by tags such as `<em>`, which is commonly used to italicize movie names.

### Details
Check if the regex used to pull the question element from the FAQ block is capturing inner tags such as `<em>`. We should also strip all tags from the question element.

### Screenshots
Screen Shot 2023-05-10 at 10 29 52 AM

jessica-townsend:
### Related Issues
- Closes #90 

### What Was Accomplished
- Changed which match from the regex is used for the question text

### How It Was Tested
- [x] Locally
- [x] Staging: https://thoughtcatalog-network-develop.go-vip.net/collectiveworld/duis-aute-irure-dolor/

### Deploy Steps
- Update Cata Blocks on thoughtnet

### DIFF
--- a/blocks/faq/includes/structured-data/class-faq-data-collector.php
+++ b/blocks/faq/includes/structured-data/class-faq-data-collector.php
@@ -35,7 +35,7 @@ class FAQ_Data_Collector {
 		$regex_question = '/<h[1-6].*wp-block-cata-faq__question.*>(.*?)<\/h[1-6]>/is';
 		
 		if( preg_match( $regex_question, $block_content, $matches ) ) {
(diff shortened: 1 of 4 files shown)


douglas-johnson (commented):
After this change, do we still need the capture group for the heading's content?

jessica-townsend:
@douglas-johnson good point, I removed that group and pushed to staging

PR #102, opened 2023-09-28
Issue#101/network blocks

### LINKED ISSUE #101: Implement a network logos block
### Background
The new design for the Latest page on Collective World has multiple areas where a logo from another site in the network is used. It would be great if we could create a block where we can add these logos and links to our sites.

### Details
- We could refer to the [social links block](https://github.com/WordPress/gutenberg/tree/5da68dc3eba8fc55214a6d4826a702663b51d295/packages/block-library/src/social-links) as a reference for implementing this one
- Note that not all logos will be the same width:
  Screen Shot 2023-09-25 at 2 49 02 PM
  How should we handle this?
  - Ensure that all SVG files are the same width
  - Have a max height along with a max width
- How should these logos be output?
  - SVG Markup
  - `` tag

### References
- https://github.com/WordPress/gutenberg/tree/5da68dc3eba8fc55214a6d4826a702663b51d295/packages/block-library/src/social-links

jessica-townsend:
### Related Issues
- Closes #101 

### What Was Accomplished
- Registered 2 new blocks - Network Link and Network Links
  - This is based off of the Social Link and Social Links block
  - Network Link blocks can only be used in a Network Links block, and Network Links blocks only accept Network Link blocks
- Created variations of the Network Link block for each site logo
- Set up SVGs for site logos with equal viewBox widths to ensure consistent sizing
- Added basic CSS for the editor to remove default `<ul>` styling and ensure that the logos have a size
- Hardcoded site URLs as an attribute for each variation

### How It Was Tested
- [x] Locally on parent theme
- [x] On child theme staging site

### How To Test
- [x] Adding a Network Links block and selecting logos to use works as expected
- [x] Markup on the frontend looks correct
  - Note that styling this block still needs to be done in the Cata theme
- [x] URLs are correct for each logo


blocks/network-link/network-link.php
```diff
@@ -0,0 +1,134 @@
...
...8.11-1.17A2.031 2.031 0 0 1 89.3-.01Zm-1.52 19.65V6.61h3.08v13.04h-3.08Zm13.76 0h-3.11L93.15 6.61h3.43l3.46 9.32 3.37-9.32h3.27l-5.12 13.04Zm18.52-3.72c-.7 2.27-2.76 4.12-5.89 4.12-3.53 0-6.67-2.58-6.67-6.99 0-4.12 3.05-6.86 6.34-6.86 4.02 0 6.37 2.65 6.37 6.77.01.34-.01.67-.08 1h-9.51c.08 1.98 1.64 3.4 3.56 3.4s2.84-1 3.27-2.27l2.62.82Zm-2.97-4.22c-.06-1.52-1.07-2.89-3.21-2.89-.8-.01-1.57.28-2.17.82s-.96 1.28-1.02 2.08h6.39ZM151.91.68h3.31l-5.33 18.98h-3.31l-4.6-14.35-4.6 14.34h-3.3L128.7.68h3.37l3.81 13.86L140.32.68h3.34l4.53 13.98L151.92.68Zm16.65 12.44c0 4.02-2.87 6.93-6.77 6.93s-6.74-2.92-6.74-6.93 2.87-6.9 6.74-6.9 6.77 2.91 6.77 6.9Zm-3.13 0c0-2.7-1.71-4.1-3.65-4.1s-3.65 1.39-3.65 4.1 1.74 4.15 3.65 4.15 3.63-1.43 3.65-4.15Zm13.21-3.43c-.33-.05-.67-.08-1.01-.07-2.42 0-3.5 1.39-3.5 3.84v6.21h-3.11V6.63h3.02v2.09c.61-1.42 2.05-2.24 3.78-2.24.28 0 .55.03.82.08V9.7Zm2.61 9.96V.28h3.11v19.37h-3.11Zm18.62-2.37c0 .8.04 1.6.13 2.39h-2.97c-.09-.56-.13-1.13-.13-1.69-.62 1.1-1.98 2-3.82 2-3.75 0-6.27-2.95-6.27-6.87s2.55-6.83 6.21-6.83c2.27 0 3.43 1.04 3.84 1.88V.28h3.02l-.02 17Zm-6.44-.03c2.04 0 3.46-1.69 3.46-4.18s-1.39-4.04-3.43-4.04-3.53 1.58-3.53 4.07 1.37 4.15 3.5 4.15Z"/></svg>',
```
douglas-johnson, line 69: I think the icons should have `width` and `height` attributes so that during page load they take up space and have their  correct aspect-ratio until the CSS is applied.

jessica-townsend: Good point @douglas-johnson, this is updated and on the CW dev site

douglas-johnson (approved):
Looks excellent! Approved.

Unless I missed a control it seems like these will be a fixed size determined by the theme. I saw it was 12em in the editor. Do you think we'll want more control over the size of the icons in the future?

jessica-townsend:
Yes right now the width will just be determined by the theme. I think controlling the icon size in the editor could be something we want in the future. We could do something like the social links and have preset sizes to select from, or we could potentially support the width setting on the network links block. It probably isn't necessary for this context, but we could either do it now or wait until the need arises

douglas-johnson:
> we could either do it now or wait until the need arises

I think it can wait until we need it on a second or third site.

jessica-townsend:
@douglas-johnson this is available on CW staging:
https://develop.collective.world/latest/

The width from the editor gets applied to the SVG's parent `<li>` element. Since the SVGs have hardcoded width and heights, we will have to style them in the parent theme to set a max width equal to the parent for the set width to be reflected on the frontend.

With the way we have the shop catalog logo set up so that the dash is overflow, we wouldn't be able to make the left aligned logo look exactly like the design. Let me know if you have any concerns about this. We could make a second SC logo with the dash inside the viewbox so we have both options
Screen Shot 2023-10-03 at 10 39 45 AM

jessica-townsend:
The parent block has margin & padding settings, which will be fine in this case since it's just the single logo

blocks/network-link/network-link.php
```diff
@@ -0,0 +1,137 @@
...
+	$width   = ( isset( $attributes['width'] ) ) ? $attributes['width'] : 'auto';
+
+	if ( ! $url ) {
+		return '';
+	}
+
+	$link  = '<li class="cata-network-link cata-network-link-' . $service . '" style="inline-size:' . esc_attr( $width ) . ';">';
+	$link .= '<a href="' . esc_url( $url ) . '" class="wp-block-cata-network-link-anchor">';
+	$link .= block_cata_network_link_get_icon( $service );
+	$link .= '<span class="wp-block-cata-network-link-label screen-reader-text">';
```
douglas-johnson, line 51: When including screen reader text, this article and the others it cites at the bottom suggest we use attributes to hide the SVG itself from screen readers. https://css-tricks.com/accessible-svg-icons/

jessica-townsend: I removed the screen reader text and added an `aria-label` attribute to the anchor tag like the article suggested. I also added `aria-hidden` to the SVGs

PR #103, opened 2023-10-12
Issue#100/rest block

### LINKED ISSUE #100: Migrate WP REST API Block from site-specific to network-wide plugin
### Background
We have a design for a new "latest" page on Collective World, which includes sections that display the latest posts from other sites on our network. We have previously implemented this capability for the Thought Catalog homepage. We could migrate that block to this plugin and add a new `Layout` class to match the design.

### REST Block
We have a custom block in the Thought Catalog repo that fetches and displays posts based on the WP REST API. These posts can be from the same site, or another site in our network ( it utilizes author data that is added by Cata Co-Authors Plus, so only sites in our network would really work).

https://github.com/wpcomvip/thought-catalog/tree/master/plugins/ought-blocks/blocks/rest

### Server Side
Substantial portions of the code have already been copied into Cata Blocks already and are used by the Products Block. The systems for querying the REST API, caching the results and updating via cron events are all used by the Products Block and are nearly identical except for the cache keys and event names. It should be possible to make those values configurable in order to use the same server-side code for both blocks.

https://github.com/wpcomvip/thought-catalog/tree/master/plugins/ought-blocks/blocks/rest/includes

### Editor 
There is also some code re-use in the editor. In order to query the REST API of another site in the editor, we needed to go through a proxy on our own site.  That is done similarly in both:
https://github.com/thoughtis/cata-blocks/blob/main/blocks/products/src/api-fetch-middleware.js
https://github.com/wpcomvip/thought-catalog/blob/master/plugins/ought-blocks/blocks/rest/src/api-fetch-middleware.js

In both cases the proxying is achieved by adding a REST API endpoint of our own. We will only need one of these, so it should probably be the version in Cata Blocks.

https://github.com/thoughtis/cata-blocks/blob/main/blocks/products/includes/proxy/class-proxy.php
https://github.com/wpcomvip/thought-catalog/blob/master/plugins/ought-blocks/blocks/rest/includes/proxy/class-proxy.php

### Styles
The styles for this block are part of the Thought Catalog 2022 theme and is spread across some block specific styles and re-usable theme styles for previewing posts.

https://github.com/wpcomvip/thought-catalog/tree/master/themes/thoughtcatalog-2022/assets/src/css/blocks/wp-block-ought-rest
https://github.com/wpcomvip/thought-catalog/tree/master/themes/thoughtcatalog-2022/assets/src/css/components/preview

The styles for the editor provides clues at to which styles are essential for the layout and which are part of the look and feel of Thought Catalog
https://github.com/wpcomvip/thought-catalog/blob/master/plugins/ought-blocks/blocks/rest/src/editor.scss

jessica-townsend:
### Related Issues
- Closes #100 

### What Was Accomplished
- Migrated the REST API block from Ought Blocks on Thought Catalog to Cata Blocks
- Set up feed and proxy PHP classes to be used by both the REST API block and the existing Products block
- Set up JS middleware to be enqueued separately and used by both the REST API and Products block
- Created a new layout called "Compact" to use in on the Collective World latest page
- Added editor styles for the Compact layout
- Removed theme-specific class names from the Network and Trending layouts

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme staging

### How To Test
- [x] Adding a REST API block and entering a URL works correctly
- [x] The Compact layout works correctly in the editor
  - This should look like a small image on the left, and a title link to the post on the right. Font sizes haven't been changed
- [x] The products block continues to work the same as before

### Deploy Steps
- Update plugin version number
- Create a new issue in the Cata theme to add basic styles for the REST API block on the frontend


includes/feed/cache/class-cache.php
```diff
@@ -59,7 +59,7 @@ class Cache {
 	 */
 	public function __construct( string $url ) {
 		$this->url  = html_entity_decode( $url );
-		$this->hash = wp_hash( $this->url );
+		$this->hash = wp_hash( $url );
```
douglas-johnson, line 62: this should go back to using `$this->url` because we want `html_entity_decode` to be applied.

includes/feed/class-feed.php
```diff
@@ -55,7 +55,7 @@ public function get_posts_allow_side_effects() : array {
 	/**
 	 * Maybe Set Update Schedule
 	 */
-	public function maybe_set_update_schedule() : void {
+	public function maybe_set_update_schedule() {
```
douglas-johnson, line 58: Is there a specific reason to get rid of the type hints here? `void` isn't terribly important but it's more accurate than nothing.

includes/feed/fetch/class-fetch.php
```diff
@@ -79,7 +70,7 @@ public function get_posts() {
 				'cata-blocks',
 				'Unable to fetch posts from remote.',
 				array(
-					'url' => $this->url,
+					'url' => $this->url
```
douglas-johnson, line 73: There are multiple changes here just for removing commas, but I think the commas are in the code standard.

includes/feed/update/class-update.php
```diff
@@ -24,7 +24,7 @@ public function __construct() {
 
 	/**
 	 * Do Scheduled Update
-	 *
+	 *
```
douglas-johnson, line 27: don't need this whitespace change

blocks/rest/src/components/post-compact/PostCompact.jsx
```diff
@@ -0,0 +1,44 @@
...
+	const dimensions = getDimensions( data );
+
+	return(
+		<article className="preview is-layout-compact">
+			<div className="preview__layout">
+				<div className="preview__start">
+				{ null !== data && (
+					<figure className="preview__image-container">
+						<a rel="bookmark" href={ post.link }>
+							<Image data={ data } dimensions={ dimensions } sizes="(max-width: 40em) 92.5vw, 36em" />
```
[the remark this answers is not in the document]
jessica-townsend: Yes, I had copied this file from `PostNetwork.jsx` and overlooked the image sizes here

includes/feed/class-feed.php
jessica-townsend: Probably better to use them, added these back

jessica-townsend:
These updates are complete @douglas-johnson, most of these changes were there because I had copied files from Thought Catalog but git is tracking them as renamed & changed from the products block

includes/feed/fetch/class-fetch.php
```diff
@@ -54,21 +54,12 @@ public function validate() : bool {
...
-			$response = vip_safe_wp_remote_request( $this->url, '', 3, 3 );
-		} else {
-			$response = wp_remote_get(
-				$this->url,
-				array(
-					'timeout' => 3
-				)
-			);
-		}
+		$response = vip_safe_wp_remote_request( $this->url, '', 3, 3 );
```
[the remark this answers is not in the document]
douglas-johnson: My dev environment for the plugin doesn't have this vip specific function. Can you copy over the if/else from the products block version where it checks if `vip_safe_wp_remote_request` exists and uses `wp_remote_get` as a back up?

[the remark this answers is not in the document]
jessica-townsend: Sorry, I missed this one before. I have this check copied over from the products block

douglas-johnson (commented):
Jess this looks great and I checked all the boxes in the PR. I think we need to make some decisions about the `.preview` class names before merging. Let's discuss tomorrow.

jessica-townsend:
Good to merge @douglas-johnson?

douglas-johnson (approved):

PR #105, opened 2023-10-28
Issue#104/shopify block

### LINKED ISSUE #104: Support Shopify in the products block
When we added a new Shopify store for Creepy Catalog products, I wrote a Cata Shopify plugin. It uses a lof the same back-end code from the products block so it already has the same Feed, Cache and Update classes as the products block.

https://github.com/wpcomvip/thoughtnet/tree/master/plugins/cata-shopify

Ways it is different:

- Requests to Shopify are done as POST requests with a GraphQL body, instead of GET requests like we've been doing for WooCommerce.
- All requests need authorization, so an API key has to be sent with each request.
  - The plugin has an options page for storing the API key and the store name.
- Because I was in a hurry and we only needed it in one place, I used a shortcode to render the content instead of a custom block. I reused the class names from the products block, but I wrote new `render_*` functions because the data from Shopify is slightly different.
- We wouldn't hit the same auth problem in apiFetch that we do with WordPress sites, but we might still want the proxy system for priming the cache and maybe for keeping the API key secret-ish.

Some of the unknowns:

- When we add a Shopify store for Shop Catalog, we'll have two stores. How do you select the store when building your block.
- We'll need something to convert the users selections in the editor ( categories, sorting, etc... ) into GraphQL.

jessica-townsend:
### Related Issues
- Closes #104 

### What Was Accomplished
- Created a block that pulls products from the Shopify API
  - Products render in the editor and the frontend
- Added settings to adjust product tags, number of products, and price visibility
- Added settings page to input information for multiple Shopify stores, which can then be used on the Shopify block
- Added a Reel style 
- Added CSS files that are used in the editor and the frontend to support basic layout & sizing

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme staging 

### How To Test
- [ ] The Shopify Products block is not rendered on the frontend under the following conditions:
  - [ ] No stores are added on the Shopify API settings page
  - [ ] A valid store hasn't been selected in the block settings
- [ ] Multiple subdomain and access token pairs can be added through the Shopify API settings page
- [ ] Subdomain and access token pairs can be removed by clearing the unwanted fields and saving the settings
- [ ] After adding a valid subdomain and access token pair, the new store is available to select in the block settings
- [ ] Shopify Products blocks with a valid store selected render products in the editor and on the frontend
  - [ ] Filtering by tag works as expected
  - [ ] Adjusting product count works as expected
  - [ ] Price visibility can be toggled
  - [ ] Reel style works as expected

### Deploy Steps
- Create a custom app to generate an access token for the new Shop Catalog store
  - https://help.shopify.com/en/manual/apps/app-types/custom-apps
- Create issues to support the Shopify Block on child themes


blocks/shopify/includes/options/store/class-store.php
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	/**
+	 * Field Callback
+	 * 
+	 * @param array $args Everything you need for an input.
+	 */
+	public static function field_callback( array $args ) : void {
+		$key = 0;
+		$setting = json_decode( get_option( self::SETTING_NAME, '' ), true );
```
douglas-johnson, line 82: I get an error here when there's no data in this setting yet.

Doing `json_decode()` on an empty string returns null, which we can't `count()`

Using `'[]'` as the default value instead of `''` works since it decodes to an empty array.

It would also work to break up this one-liner so it can be type checked before trying to decode and count.

blocks/shopify/includes/options/store/class-store.php
```diff
@@ -0,0 +1,124 @@
...
+	 * @return array
+	 */
+	public static function sanitize_settings( $array ) : string {
+		if ( empty( $array ) ) {
+			return array();
+		}
+
+		$array = map_deep( $array, 'sanitize_text_field' );
+
+		$array = array_filter( $array, function( $value ) {
```
douglas-johnson, line 68: I'm getting an error here.

The error says that a string value was provided to array_filter.

I logged the value of $array, and it is being logged twice.

The first time it is an array. The second time it is a json_encoded string containing the same data.

So for some reason this is being run twice, and the second time, the data is already encoded.

blocks/shopify/includes/options/store/class-store.php
jessica-townsend: I think `json_decode` only accepts a string, so `[]` being the default option value didn't work for me
https://www.php.net/manual/en/function.json-decode.php

But I can still check if the option is valid json before continuing

blocks/shopify/includes/options/store/class-store.php
jessica-townsend: Did this only happen the first time you saved something? I'm wondering if it's related to this:
https://core.trac.wordpress.org/ticket/21989

It would probably be best to modify the data with this filter before it gets saved instead
https://developer.wordpress.org/reference/hooks/pre_update_option_option/

blocks/shopify/includes/options/store/class-store.php
jessica-townsend: Oh gotcha, I misunderstood!

blocks/shopify/includes/options/store/class-store.php
douglas-johnson: I think you're right about that issue, but I think we can solve it with less code by using the WordPress core's ability to save an option as an array. I think it uses [serialize](https://www.php.net/manual/en/function.serialize.php) instead of JSON.

I tried it out. Here are the changes I did to use a serialized array instead JSON. I'll send you a patch file with the same changes for easy copying.

class-store.php
Screenshot 2023-10-30 at 1 27 55 PM
Screenshot 2023-10-30 at 1 28 02 PM
Screenshot 2023-10-30 at 1 28 20 PM


Screenshot 2023-10-30 at 1 29 20 PM

jessica-townsend:
@douglas-johnson I've applied the patch that you sent and that worked great. However, I started getting this warning:
Screen Shot 2023-10-30 at 11 17 05 AM

I read this about adding a schema:
https://make.wordpress.org/core/2019/10/03/wp-5-3-supports-object-and-array-meta-types-in-the-rest-api/

I'm unsure how the schema should be formatted since the data is a nested array. Everything continues to work properly without a schema added besides the warning. I tried the following which all cause the settings API to return `null` for the "Stores" dropdown in the block settings:
```
'show_in_rest'      => array(
	'schema' => array(
		'items' => array(
			'type'  => 'array',
			'items' => array(
				'type' => 'string',
			),
		),
	),
),

'show_in_rest'      => array(
	'schema' => array(
		'items' => array(
			'type'  => 'array',
		),
	),
),

'show_in_rest'      => array(
	'schema' => array(
		'items' => array(
			'type'  => 'string',
		),
	),
),
```

Do you have an idea for how this should be handled?

douglas-johnson:
I think I figured this out, but then I had a realization. In order for people to access `/wp/v2/settings/` they have to be able to manage options and that means being an admin. We will want editors, authors and contributors to be able to see this content in the editor.

Last time I checked, there wasn't a way to pass arbitrary settings to the `core/block-editor` store. https://developer.wordpress.org/block-editor/reference-guides/data/data-core-block-editor/

We can:
- Add our own API route that returns the store names to anyone logged in.
  - Pro: Easier to set up
  - Con: Extra HTTP request for every block ( unless we put our cache middleware on it )
- Add a custom redux store and use wp_localize_script to make the initial state available in the window.
  - Pro: No http requests
  - Con: More steps to set up

Here's an example of the custom store.

- script for the store, uses window for inital state
  - https://github.com/thoughtis/Co-Authors-Plus/blob/issue%23940/coauthor-blocks/src/blocks-store/index.js
- enqueuing the script, server side data to the window. 
  - https://github.com/thoughtis/Co-Authors-Plus/blob/issue%23940/coauthor-blocks/php/blocks/class-blocks.php#L123
- In a block, using useSelect to get the store
  - https://github.com/thoughtis/Co-Authors-Plus/blob/issue%23940/coauthor-blocks/src/blocks/block-coauthors/edit.js#L63-L66

jessica-townsend:
Thanks for the links @douglas-johnson, this is up to date on develop. I tested locally and was able to select a store in the block editor as a contributor

douglas-johnson:
React wants each Product in Products to have a key. Adding the product id to the GraphQL query and using that might be a possibility. You might also [add the index to the parameters in the map function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#parameters) and build a key like `product_${index}`

Screenshot 2023-10-31 at 1 45 26 PM

jessica-townsend:
@douglas-johnson These changes are available on staging

blocks/shopify/src/style.scss
```diff
@@ -0,0 +1,49 @@
+/**
+ * The following styles get applied both on the front of your site
+ * and in the editor.
+ *
+ * Replace them with your own styles or remove the file completely.
+ */
+
+// .wp-block-cata-shopify {}
```
douglas-johnson, line 8: There are some commented out class names here and then there are styles for them later on.

blocks/shopify/includes/options/store/class-store.php
```diff
@@ -0,0 +1,124 @@
...
+				$access_token = is_array( $option ) && array_key_exists( 'access_token', $option ) ? $option['access_token'] : '';
+				echo $key > 0 ? '<br>' : '';
+				?>
+					<fieldset id="<?php echo esc_attr( self::SETTING_NAME . '_' . $key ); ?>" data-key="<?php echo esc_attr( $key ); ?>">
+						<p><strong>Store <?php echo esc_html( $key + 1 ); ?></strong></p>
+						<br>
+						<label for="<?php echo esc_attr( $args['id'] . '_subdomain_' . $key );?>" style="min-width: 8em;">Subdomain</label>
+						<input 
+							id="<?php echo esc_attr( $args['id'] . '_subdomain_' . $key ); ?>" 
+							name="<?php echo esc_attr( $args['name'] . '[' . $key . '][subdomain]' ); ?>"
```
douglas-johnson, line 100: This all works, but I think it could be made more legible by setting id and name to variables and using double quote strings.

`name="<?php echo esc_attr( $args['name'] . '[' . $key . '][subdomain]' ); ?>" `

could change to 

`name="<?php echo esc_attr( "{$name}[{$key}][subdomain]" ); ?>"`

blocks/shopify/includes/options/store/class-store.php
```diff
@@ -0,0 +1,124 @@
...
+		$num_options = count( $setting );
+
+		?>
+			<div id="<?php echo esc_attr( self::SETTING_NAME ) . '_wrapper'; ?>" style="margin-bottom: 2em;">
+		<?php
+			do {
+				$option = array_key_exists( $key, $setting ) ? $setting[$key] : array();
+				$subdomain = is_array( $option ) && array_key_exists( 'subdomain', $option ) ? $option['subdomain'] : '';
+				$access_token = is_array( $option ) && array_key_exists( 'access_token', $option ) ? $option['access_token'] : '';
+				echo $key > 0 ? '<br>' : '';
```
douglas-johnson, line 92: Could this conditional line-break be removed if the `fieldset` elements here and in the JS for add-store all had `style="margin-bottom: 2em;"` instead of the wrapper div?

blocks/shopify/src/style.scss
jessica-townsend: Thanks, I must have overlooked this after I copied from the products block

blocks/shopify/includes/options/store/class-store.php
jessica-townsend: Yes using margin instead works here!

douglas-johnson (approved):

PR #107, opened 2023-11-02
Issue#106/support multiple tags

### LINKED ISSUE #106: Filter products from Shopify API by author
### Background
We added a new block to display products from any Shopify site given a subdomain and access token. We will likely need to filter by multiple tags, specifically to get products by a certain author.

### Details
- If Shopify doesn't allow commas in tag names, we could use a comma separated list in the current Tag input
- If Shopify does allow commas in tag names, then we can set up another input for an author

jessica-townsend:
### Related Issues
- Closes #106 

### What Was Accomplished
- Added support for multiple tags in a comma separated list
- Added CSS to make all product images have the same aspect ratio

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Staging child theme

### How To Test
- [x] Adding multiple tags in the input field as a comma separated list only pulls products from Shopify that have both of those tags
  - This should work with or without spaces between the comma and next tag name
- [x] Tags with spaces in the tag name continue to work as expected
- [x] Only adding a single tag continues to work as expected
- [x] No added tags continues to work as expected
- [x] Images are consistent sizes

### Deploy Steps
- [x] Update plugin version number
- [ ] Make issues to update plugin on other sites


douglas-johnson (approved):
Approved, go ahead and merge this, we can can take it to the live sites on Monday. Great work!

PR #124, opened 2024-01-17
Issue#123/decode rest urls

### LINKED ISSUE #123: Decode REST API block URLs
### Background
We've noticed an issue with the Horoscope layout on the REST API block that's causing the output to "freeze" and no longer update to the latest horoscope post. The horoscope links don't work either. The editor shows ~10 posts with different dates. The REST API URL in the editor has `&amp;` instead of `&` for the query string, which causes any `tag` or `per_page` arguments to not work correctly. It seems like the `&` symbol gets changed if another user edits the page.

### Details
- A similar issue was fixed on the products block: https://github.com/thoughtis/cata-blocks/issues/36
- We should try decoding the URL in JS before pushing it into the URLs array, and before saving the URLs array to the block attributes

jessica-townsend:
### Related Issues
- Closes #123 

### What Was Accomplished
- Decode HTML from API URLs on the edit screen for the REST block when a URL gets added and when posts are retrieved.

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme develop

### How To Test
- On the local main branch, create a REST API block and add a url with query parameters using `&amp;` instead of `&`. The displayed posts should ignore any query parameters.
- Switch to this branch and update the page with the REST API block without making any changes. The posts displayed in the editor and on the frontend should update to honor the URLs query parameters. The URL listed in the editor will still contain `&amp;`
- Test adding another URL to the REST API block that contains `&amp;`. It should be replaced with `&`.


blocks/rest/src/edit.js
```diff
@@ -95,6 +95,12 @@ export default function Edit( { attributes, setAttributes, clientId } ) {
 		});
 	}
 
+	function decodeHtml(text) {
```
douglas-johnson, line 98: I thought surely there must be an easier way, but after looking it up I see why this is popular. Do either of the options below strike you as less code in any useful way? If you want to stick with the current version, could you add a link or some function documentation about why we're doing this?

Since lodash comes with WordPress, we could also use `_.unescape` 
https://lodash.com/docs/4.17.15#unescape

or if we're really only concerned with `&amp;` it could be solved with `.replaceAll('&amp;', '&')`

jessica-townsend: Thanks! I decided to switch to `_.unescape` which still seems to work fine. I just pushed to develop again too.

douglas-johnson: One more thing about this, I feel lodash should be listed in the dependencies. 

If you add `import { unescape } from 'lodash'` at the top, does `wp-scripts` add it to the dependencies array?

jessica-townsend: It does @douglas-johnson, just added that to `edit.js`

douglas-johnson (approved):

PR #127, opened 2024-01-22
Issue#126/newsletter block

### LINKED ISSUE #126: Update newsletter block to be used across multiple themes
### Background
The newsletter block has default values and styles that were meant specifically for the Creepy Catalog theme. We would like to use this block on other themes, so it needs an update to make it more generic.

jessica-townsend:
### Related Issues
- Closes #126 

### What Was Accomplished
- Removed background image, title, and description attributes from the Newsletter block
- Removed theme-specific styles
- Moved some layout styles to a combined frontend and editor CSS file
- Adjusted HTML structure for the block output
- Added support for border and spacing settings in the editor
- Updated theme-specific default attribute values

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme develop


douglas-johnson (changes requested):
This is really great, there's just even more that can be removed.

Unless there's a reason I'm missing, I think you could:
- Remove the `.wp-block-cata-newsletter__inner` element 
- Remove `a { color: currentColor; }` because we do that in all the themes now
- Use `get_block_wrapper_attributes()` instead of our custom `Renderer::get_classes`
- search for `$atttributes` and correct the spelling

jessica-townsend:
Thanks @douglas-johnson, I have all of that updated

blocks/newsletter/includes/class-renderer.php
```diff
@@ -35,11 +35,11 @@ public function __construct( array $block_attributes ) {
...
 	public function get_content() : string {
-		$form        = $this->get_form();
-		$legal_text  = $this->get_legal_text();
-		$classes     = $this->get_classes(
+		$form         = $this->get_form();
+		$legal_text   = $this->get_legal_text();
+		$wrapper_attr = get_block_wrapper_attributes(
 			array(
-				'wp-block-cata-newsletter',
+				'class' => 'wp-block-cata-newsletter',
```
douglas-johnson, line 42: Does it automatically produce the same class name that we're assigning here?

jessica-townsend: Yep it does, removed

douglas-johnson (approved):
This looks good to me, I tried it out in the Thought Catalog theme.

PR #131, opened 2024-02-08
Issue#122/flex grow control

### LINKED ISSUE #122: Add custom flex-grow control
The controls for the flex layout type allow, fit, fill and fixed sizes. We really need fixed and fill to work together. We run into multiple layouts that would benefit from support for uses cases like "20em by default and grow to fill the available space".

- Can we add control to the layout panel to set flex-grow as a number?
- Based on that, flex-grow be set in the style attribute of a block?

jessica-townsend:
### Related Issues
- Closes #122 

### What Was Accomplished
- Added a `cataBlocksFlexGrow` attribute with a number control that is only available if a block is fixed width

### How It Was Tested
- [x] Local parent theme
- [x] Local child theme
- [x] Staging child theme

### How To Test
- Flex grow number control appears for all blocks that are a direct child of a block with a `flex` layout and with the width set to `fixed`
- The `flex-grow` style for a block appears in the editor and on the frontend
- The `flex-grow` style is removed and the number control field is nulled if the block is switched to `fit` or `fill`


douglas-johnson (changes requested):
It works great in my local testing. I think these could also use `import` instead of `wp.`

const { __ } = wp.i18n;
const { __experimentalNumberControl: NumberControl } = wp.components;
const { InspectorControls } = wp.blockEditor;

douglas-johnson (approved):

PR #133, opened 2024-03-06
Support customClassName in Table of Contents Block

### LINKED ISSUE #132: Using styles requires passing className to useBlockProps.save()
We want to use block styles on the table of contents block. Currently, registered styles work in the editor but are not saved or output in the rendered content.

The problem is resolved if we updated the block's save function to include `...useBlockProps.save({ className: attributes.className })`

douglas-johnson:
### Related Issues
- Resolves #132 

### What Was Accomplished
- Updated the Table of Contents block's block.json to remove declared non-support for customClassName. Support for customClassName is required for the Block Styles API.

### How It Was Tested
- [x] Local plugin dev environment
- [x] Local theme dev environment with a registered style.

### How To Test
- [ ] Applying a custom CSS class to the table of contents block in the editor persists to the rendered content.


jessica-townsend (approved):
Works for me locally

PR #145, opened 2024-04-12
Issue#142/cata blocks styles

### LINKED ISSUE #142: Add block styles from Cata to Cata blocks
### Background
We've been using the Cata theme to style blocks in Cata Blocks, and we've run into several occasions where it would have been better if the basic styles for our custom blocks were in Cata Blocks. We should copy the CSS for our custom blocks from Cata to Cata Blocks so that we can easily use Cata Blocks on sites that don't use the Cata theme.

### Resources
https://github.com/thoughtis/cata/tree/master/assets/src/less/blocks

jessica-townsend:
### Related Issues
- Closes #142 

### What Was Accomplished
- Copied block styles from the Cata theme to the Cata Blocks plugin

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme dev site

### How To Test
- Basic styles for all of our custom blocks are present, even if Cata theme styles are disabled

### Deploy Steps
- We will need to remove any redundant styles from the Cata theme and out Thought Catalog themes


douglas-johnson (changes requested):
I am requesting changes, but mostly because I want to review more thoroughly. I can see this working locally on multiple sites when I manually turn other styles off, but I'd like to see a complete version when I don't have to change anything.

Could you rebase this branch and set up the branch for Cata where the duplicate styles are removed, and then I can review them both?

jessica-townsend:
Yes, I can do that @douglas-johnson !

douglas-johnson (approved):

PR #151, opened 2024-05-24
Issue#150/marquee block

### LINKED ISSUE #150: Marquee block
### Background
The site designs for Agency includes a section where heading text will scroll across the page:
Screenshot 2024-05-22 at 2 44 20 PM

We need to implement this as a block. It would be best if the block could have inner content that accepts headings and paragraphs, since this effect will be used at the bottom of the page too. Here's a codepen that shows the markup structure:
https://codepen.io/jessicaat123/pen/ExzygeE

Note that we will also need styles for `prefers-reduced-motion`

jessica-townsend:
### Related Issues
- Closes #150 

### What Was Accomplished
- Created a Marquee block that contains a Post Title block and will scroll the text across the screen

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme develop

### How To Test
- Animation is removed if the browser has `prefers-reduced-motion`
- There is no "jumping" in the animation on various screen sizes
- The animation works properly for various text sizes

### Notes
- I changed the pseudo element to be `absolute` with extra padding on the parent element which I felt helped reduce the blank space between text on smaller screens
- I added an `!important` flag to left side padding because this was being added by core, causing the animation to jump:
  Screenshot 2024-05-24 at 11 19 03 AM
  However I don't see this style being added on the Agency theme.


douglas-johnson:
I tested this on the development site and everything seemed to work except the font-size in the rendered content is different form the editor I don't understand why.

https://develop.thoughtcatalog.network/agency/case-studies/influencer-production/

I'm getting `has-step-p-3-font-size` even though I didn't set that in the editor.

Screenshot 2024-05-28 at 11 53 35 AM

jessica-townsend:
@douglas-johnson the marquee block at the very top comes from the case study template, not from individual case study post content. Any blocks added to a case study post appear below the featured image

douglas-johnson:
GOT IT! My bad on that one. Next thing I'm seeing is I put alignfull on it, which isn't working.

jessica-townsend:
@douglas-johnson I think that's because the content block was inside a group on the case study template. I moved it out of the group and now the marquee block is full width on my end

douglas-johnson (approved):

PR #160, opened 2024-07-10
Issue#159/globe background block

### LINKED ISSUE #159: Animated globe background block
### Background
The new design for the Agency homepage includes a network section that is similar to the animated globe on the current site.

### Details
- After attempting to use a custom style on a group block, I think it would be best to just use a new block for this animation. It will be easier to adjust settings and display the animation on the frontend and a background in the editor

### Attachments
Screenshot 2024-07-09 at 9 06 46 AM

jessica-townsend:
### Related Issues
- Closes #159 

### What Was Accomplished
- Added a container block with an animated globe background on the frontend

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme develop

### How To Test
- Globe stroke color can be changed in the editor and is reflected on the frontend
- Globe is behind content in the editor and on the frontend
- Globe is animated on the frontend (not in the editor)
- On smaller screens, globe does not cause a horizontal scrollbar


douglas-johnson (changes requested):
Every item in "how to review" checks out, but the prefers-reduced-motion version leaves a little to be desired. On the current agency site if you turn on reduced motion, you'll see the globe kind of frozen in place. In this new version I'm only getting one big circle.

douglas-johnson (approved):
The update to the reduced motion works for me locally. Thanks!

PR #170, opened 2024-07-23
Display daily horoscope links in REST API block's network layout

### LINKED ISSUE #168: Optionally include tarot links in network layout?
We want to be able to achieve both of these looks:

The current one where horoscopes can be dropped in to a small space and its possible to add content around them.

Screenshot 2024-07-22 at 3 06 23 PM

Another that looks more like the network article layout but with the addition of the date and horoscope links.
hppatterns (1)

douglas-johnson:
### Related Issues
- Resolves #168 

### What Was Accomplished
- Updated JS and PHP in REST API block so the daily horoscope links can be displayed in the network layout.
- Updated the CSS to vertically center the zodiac signs without inline-flex ( which creates empty space at the bottom ).

### How It Was Tested
- [x] Local
- [x] Remote

### How To Test
- [x] Horoscope links appear in content using the Network layout. They are always visible in the editor, but only present in the rendered content when they actually exist.
- [x] Horoscope links still appear as normal in the Daily Horoscope layout. 
- [x] Network layout without horoscopes works as normal.


jessica-townsend (approved):

douglas-johnson:
@jessica-townsend I'm double pointing out that `display_zodiac_links` now defaults to false so when this ships to a site we have to go flip the toggle where we want the links to show.

jessica-townsend:
Thanks for reiterating @douglas-johnson, I'll make note of that

PR #171, opened 2024-09-11
Issue#123/decode rest urls

### LINKED ISSUE #123: Decode REST API block URLs

jessica-townsend:
### Related Issues
- Closes #123 

### What Was Accomplished
- Removed instances where REST URLs are decoded before saving to the database
- Decode REST URLs later in the process, before they are used

### How It Was Tested
- [x] Locally on parent theme
- [x] Staging child theme

### How To Test
- New REST blocks where the URL contains an HTML entity (`&amp;`) still display the correct posts in the editor and on the frontend
- Existing REST blocks are unchanged (besides horoscope links being hidden by default)
- REST blocks where the URL contains an HTML entity are unchanged after the `cata_blocks_feed_update` cron event runs

### Deploy Steps
- Note that the horoscope link toggle is set to false by default now, so we need to update any current REST blocks in the editor that display horoscope links after deploying.


douglas-johnson (approved):
Saw this work in the editor and rendered cotent locally and watched the cron be executed on the development site.

PR #172, opened 2024-09-25
Issue#59/shopify block aspect ratios

### LINKED ISSUE #59: Support portrait and landscape feature image sizes
### Background

On Shop Catalog we have an established pattern for feature images that should allow the vast majority of products to support portrait, landscape and square version of their feature image. The products block is all square because of the way we designed it for Creepy Catalog where it has a film strip frame around it.

We recently added the reel style and the square images make the final output a little larger than necessary. If we were able to use the portrait crop, the resulting image would contain more of the product, especially in the case of books.

That could mean we can fit more products on the page horizontally in the reel layout.

### Details

- A user should be able to select between square, portrait and landscape image orientations.
- Square is currently hard coded into the JS and PHP versions of the output
  - https://github.com/thoughtis/cata-blocks/blob/0.7.1/blocks/products/src/components/product.js#L24-L34
  - https://github.com/thoughtis/cata-blocks/blob/0.7.1/blocks/products/includes/renderer.php#L49-L59
  - https://github.com/thoughtis/cata-blocks/blob/0.7.1/blocks/products/includes/renderer.php#L101
- The image sizes from Shop Catalog are listed below.
- Because Shop Catalog is using the WordPress VIP CDN, we can dynamically crop images using URL parameters. Updating the image URLs with new dimensions should work automatically.

```php
// 3:4
add_image_size( 'sc_portrait_sm', 384, 512, true );	
add_image_size( 'sc_portrait_md', 768, 1024, true );
add_image_size( 'sc_portrait_lg', 1152, 1536, true );

// Square
add_image_size( 'sc_square_sm', 384, 384, true );	
add_image_size( 'sc_square_md', 768, 768, true );
add_image_size( 'sc_square_lg', 1152, 1152, true );

// 3:2
add_image_size( 'sc_landscape_sm', 384, 256, true );	
add_image_size( 'sc_landscape_md', 768, 512, true );
add_image_size( 'sc_landscape_lg', 1152, 768, true );
```

jessica-townsend:
### Related Issues
- Closes #59 

### What Was Accomplished
- Added select field to Shopify block to choose an aspect ratio for the images
- Adjusted image dimensions in the Shopify CDN URL based on selected aspect ratio

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme develop ( Thoughtnet )

### How To Test
- Images from the Shopify block show the selected aspect ratio in the editor and in the rendered content
- Images from the Shopify block have correct CDN URLs based on aspect ratio in `srcset` and `src`


douglas-johnson:
I think I'm hitting some scenarios where the images in the `srcset` are correct but the `src` is not. I was seeing some warping when I used the list layout.

Screenshot 2024-09-25 at 3 08 48 PM

jessica-townsend:
@douglas-johnson I think this is happening because it's currently only comparing the images target width with the original image width and not the height, so we are trying to support a height that's taller than the original image when using a portrait aspect ratio. I'll do some adjusting so we check the height as well.

jessica-townsend:
This is updated on develop

PR #185, opened 2025-01-07
Issue#141/flex grow post blocks

### LINKED ISSUE #141: Figure out why the new flex-grow attribute doesn't work on some blocks
### Background
We've noticed that on some blocks, the flex-grow attribute doesn't appear on the frontend. We need to investigate where this happens and why. There are a few possibilities:
- Inside a query loop block
- On image blocks
- On blocks that already have a `style` attribute

jessica-townsend:
### Related Issues
Closes #141 

### What Was Accomplished
- Added a filter for post featured image blocks that adds our custom `flex-grow` attribute to the inline `style` tag

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme develop

### How To Test
- [x] Add a featured image block to a row block. Then set a fixed width and flex grow in the featured image block styles.
- [x] The flex grow value should be reflected in the editor and the frontend.
- [x] Other inline styles such as margin still work properly on the featured image block.

### Open Questions
This is the only `post-` block that I felt needed to have flex grow work. I think like it makes sense to keep wrapping text elements such as excerpt and title blocks in a group if we need to set flex grow. We could remove the flex grow field in the editor for these other blocks, but we can make a new issue for that if it becomes confusing in the future

### DIFF
--- a/cata-blocks.php
+++ b/cata-blocks.php
@@ -12,7 +12,7 @@
  * Description: Block Editor components for use with the Cata theme.
  * Author:      Thought & Expression Co. <devjobs@thought.is>
  * Author URI:  https://thought.is
- * Version:     0.10.11
+ * Version:     0.10.12-beta1
  * License:     GPL v3 or later
  * License URI: http://www.gnu.org/licenses/gpl-3.0.txt
  */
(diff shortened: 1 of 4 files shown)


douglas-johnson (approved):
This is working great, thanks for figuring this out!

PR #189, opened 2025-02-06
Issue#188/image aspect ratio

### LINKED ISSUE #188: Add aspect ratio controls to the REST API block
### Background
Now that we have more use cases for this block, it would be great if we could change the aspect ratio of the images to have more flexibility.

### Details
- The Shopify block uses some custom aspect ratio options
- You can now declare support in `block.json` for `aspectRatio` as of WP 6.5, however this will be applied to the parent element and not the images directly. We could probably move the classnames and styles to the images in our render callback function.

jessica-townsend:
### Related Issues
- Closes #188 

### What Was Accomplished
- Added an aspect ratio setting to the REST block on the Network layout

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme develop

### How To Test
- [x] Aspect ratio selector is available under "Layout" only if if the default or Network layout is selected
- [x] The selected aspect ratio is reflected on the frontend and in the editor
- [x] No aspect ratio is used if the layout is changed or the aspect ratio is set to "Original"


douglas-johnson (approved):
Working great on develop

Issue #190, opened 2025-02-21
Even more compact REST API block


douglas-johnson:
### Background

We'd like to display some small sets of posts from our other sites above and below the post content. The current compact layout is pretty small but the inclusion of the images makes it still a bit large for this context where we won't want to totally distrupt the flow of reading.

### Details

- Add a new layout, most likely something similar to the right column of the featured layout which is just text links in paragraphs

PR #193, opened 2025-03-10
Make all arguments after $posts in rest block render functions optional

douglas-johnson:
### Related Issues
- #190 
- After shipping #191 I encountered "too few arguments" errors in environments where the aspect ratio had never been set for the compact layout.

### What Was Accomplished
- Updated all `render` functions in the REST API block layout classes so that the arguments after `$posts` are optional and have default values that make them have no effect.

### How It Was Tested
- [ ] Local
- [ ] Remote

### How To Test
- [ ] Existing blocks using compact layout continue to work
- [ ] No `too few arguments` errors
- [ ] Blocks update over time ( errors result in old content from when the block was added to the post being displayed )

### DIFF
--- a/blocks/rest/includes/layout/compact/class-compact.php
+++ b/blocks/rest/includes/layout/compact/class-compact.php
@@ -23,7 +23,7 @@ class Compact extends Layout {
 	 * @param array $posts
 	 * @return string
 	 */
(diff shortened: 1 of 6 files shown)


jessica-townsend (approved):
Looks good to me on develop

PR #194, opened 2025-03-11
Issue#192/advanced navigation block

### LINKED ISSUE #192: Advanced navigation block
### Background
Since starting to use the full site editor, creating custom navigations has been a limitation for us. We would like to create modal navigations where we can add social links, static text, search bars, and multiple menus. Additionally, we want the modal to have a backdrop against the visible content. None of this is supported by WP core, so it would be great if we could make a more advanced navigation block

### Details
- Make a parent "modal block", which will be a hamburger button with a dialog element that accepts `innerBlocks`.
- Make a child "close modal" block in the dialog element's `innerBlocks` that can be moved but not removed.
- How does `multiple: false` on a custom block works if that block is inserted into a template part? It would be great if we can only have one modal per page. Otherwise we may need a way to set a unique ID to target opening and closing the dialog.

jessica-townsend:
### Related Issues
- Closes #192 

### What Was Accomplished
- Added an advanced modal block that can be used to create more complex dialogs through the editor
- Added a separate close button block so its position in an open modal can be changed

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme develop

### How To Test
- [x] Modal can be opened in the editor and on the frontend
- [x] The backdrop color & opacity can be changed in the editor and is reflected on the frontend
- [x] The color of the default hamburger icon can be changed in the editor and is reflected on the frontend
- [x] The color of the close button is inherited from its parent
- [x] The dialog element has a scrollbar if the content overflows
- [x] Multiple dialogs can be added to a page, and each dialogs related buttons control their associated dialog

### Open Questions
- We've discussed having multiple buttons control a single dialog. If we implement that feature, how can we set what dialog the button controls? Or should we switch to only allowing a single dialog per page?
- I felt like the close button block behaved strangely when trying to move it around within it's parent. I think its because its set to only be added as a child of the advanced modal block. We could fix this, but then it could be shown as available for use anywhere
- Outside of the template editor, the WordPress sidebar will overlap the open dialog. I had tried fixing this based on some media queries and SCSS variables that the navigation block uses, but then it wouldn't appear properly when using the template editor. This is something we could look into further if needed, but might be unnecessary since these blocks are specifically for block themes


jessica-townsend:
@douglas-johnson this is now on develop

douglas-johnson (approved):
This is awesome, worked great on develop and locally with the twenty twenty five theme

PR #197, opened 2025-03-20
Add Widget ID attribute to Outbrain block

### LINKED ISSUE #195: Add Outbrain Block
### Background

We want to add outbrain on one of our sites that is using the full site editor, so we'll need a block for it.

### Details

- Block should show placeholder content in the editor
- Block will need to include the URL of current post being viewed so it needs post context and/or a render callback that can get the URL.
- https://developer.outbrain.com/outbrain-javascript-implementation-guide/

douglas-johnson:
### Related Issues
- Resolves #195 
- After seeing this working in production yesterday the next step is to add some control so the block is reusable.

### What Was Accomplished
- Added an attribute for Outbrain Widget ID, this controls which version of the Outbrain content recommendation widget is used.
- Updated the render function to support the widget ID attribute as well as the standard block CSS classes.

### How To Review
Currently the outbrain widget itself only works in production, but you can see the block doing what its supposed to in local or remote dev environments.

- [x] If you update the Widget ID in the editor it is refelected in the `data-widget-id` attribute of the rendered content.


jessica-townsend (approved):
Tested locally

PR #201, opened 2025-04-03
Issue#200/svg links

### LINKED ISSUE #200: Add link support to the Safe SVG block
### Background
We've been using the Safe SVG block for some of our sites, and it works well to handle iconography. Some more recent site designs include hover states for logos, which would need to be SVGs to change the color on hover. By default, the Safe SVG block doesn't support links.

jessica-townsend:
### Related Issue
- Closes #200 

### What Was Accomplished
- Added support for the link toolbar setting on the inline svg block
- Added support for link colors on the inline svg block

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme develop

### How To Test
- [ ] Link toolbar setting is available in the editor for inline SVG blocks
- [ ] Link color settings is available in the editor for inline SVG blocks
- [ ] Anchor element appears on the frontend for inline SVG blocks that have a link added in the editor
- [ ] Inline SVG links open in a new tab if specified
- [ ] Inline SVG links have correct color states


douglas-johnson:
@jessica-townsend Can you send me one SVG that you're testing with? I'm not getting the color states but I bet that has to do with the SVG or related CSS.

douglas-johnson (approved):

PR #207, opened 2025-04-25
Issue#206/update dependencies

### LINKED ISSUE #206: Update dependenceis to WP 6.8 versions
### Background
Now that WP 6.8 is released, we should re-build and test our blocks using the `@wordpress` node packages that are tagged `wp-6.8`.

It looks like the `experimentalLinkControl` component is no longer experimental, so we can change that to just `LinkControl` here:
https://github.com/thoughtis/cata-blocks/blob/main/block-editor/src/components/svg-url.js#L17

jessica-townsend:
### Related Issues
- Closes #206 

### What Was Accomplished
- Updated Cata Blocks dependencies to `wp-6.8` versions
- Rebuilt all blocks
- Fixed deprecation warnings in the console
- Replaced `__experimentalLinkControl` component with `LinkControl`

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child theme
- [x] Child theme develop

### How To Test
- [x] All blocks continue to work as expected in the editor and on the frontend
- [x] Block filters continue to work as expected in the editor and on the frontend (flex grow setting and inline SVG link)
- [x] There are no errors and minimal deprecation warnings in the console in the editor related to our blocks
  - There are a couple related to the `__nextHasNoMarginBottom` and `__next40pxDefaultSize`, but it doesn't seem like there's anything we need to do for those. It sounds to me like they will just become the default for some components in a future version.  

### Deploy Steps
This should **not** be deployed to our production sites until we've updated to WP version 6.8


douglas-johnson (approved):
Everything's working great. There was an errant console.log in `/block-editor/build/index.js` but its been removed from the source so rebuilding that file would fix it.

jessica-townsend:
@douglas-johnson I think that has already been rebuilt, I see the `console.log()` was removed from the `build/index.js` file

PR #223, opened 2025-11-11
Color Scheme ( Light / Dark mode ) settings and preview

douglas-johnson:
### Related Issues
- I didn't write an issue for this one and I should have. We have a dark theme on Creepy Catalog but our writers want to use the editor in light mode.
- We are addressing this by using `light-dark()` for the color palette and making it possible to preview dark mode in the editor.
- It will also be possible to assign color modes to specific blocks, as long as they support colors in general. 

### What Was Accomplished
- Added a control in the color group that lets you select Original, Light or Dark color schemes
- Added a custom item to the preview menu in the editor to preview the entire editor canvas in a color scheme
- Added state management that figures out if the active element in the editor is using the dark color scheme and adjusted the available color palette accordingly.

### Open Questions
- "Original" might be the wrong term for inheriting the color scheme from the parent
  - I changed this to Default instead. 
- The dropdown for selecting the color scheme could use a component that more closely matches the rest of the color selection UI.
  - Could not find a way to make the select box prettier, and using the WP color selection component doesn't really make sense since we don't want to open a color palette. This might be as good as it can be for now.
  
 ### How To Test

Theme must opt in to `cata_blocks_theme_supports_color_scheme`  and use the light-dark() CSS function in the color palette. The Creepy Catalog block theme currently on develop site meets there requirements.

- [x] Individual blocks can be assigned light mode or dark mode.
- [x] The color palette available in those blocks changes according to the color assignment.
- [x] The post / page editor can preview the entire document in dark mode.
- [x] The site editor can preview the entire canvas in dark mode.


block-editor/color-scheme/src/components/color-scheme-control.js
```diff
@@ -0,0 +1,193 @@
+/**
+ * Color Scheme Control
+ */
+import {  } from '@wordpress/components';
```
jessica-townsend: Do we need this empty import?

jessica-townsend (approved):

block-editor/color-scheme/src/components/color-scheme-control.js
douglas-johnson: almost certainly not, sorry i missed that

douglas-johnson: This should be fixed in the changes i just pushed

jessica-townsend: Great, thank you! Everything else looked good to me on develop

PR #225, opened 2026-02-04
Infinite Scroll as Blocks

douglas-johnson:
### Related Issues
- We needed to load an additional post after the content on a specific site that uses the full site editor. We had code to work with from infinite scroll on hybrid theme sites. In this case we've only been asked to load one post so this code only serves that need. It could be expanded to do more if requested later.

### What Was Accomplished
- Added Infinite Scroll and Infinite Scroll Beacon blocks
  - Infinite Scroll uses InnerBlocks and wraps around the content you want to be inserted on scroll.
  - Infinite Scroll Beacon uses the IntersectionObserver to trigger the loading of content when it enters the screen.
- Added an admin settings page
  - Activate / Deactivate infinite scroll
  - Select the post being used for infinite scroll

### How It Was Tested
- [x] Local
- [x] Remote theme dev environment
- [x] Currently running in production environment

### How To Test
- [x] By inserting the Infinite Scroll and Infinite Scroll Beacon blocks adjacent to oneanother in the single post template and selecting a post through the plugin settings it is possible to load new content on scroll.

### Future Work / Open Questions
- I did not add special handling for cases where the beacon is on screen immediately on load.
- If the user scrolls back to the previous content we do not replace the state so the URL remains the same.


blocks/infinite-scroll/build/infinite-scroll-beacon/view.js
jessica-townsend, line 1: Can this file be removed?

jessica-townsend (approved):

blocks/infinite-scroll/build/infinite-scroll-beacon/view.js
douglas-johnson: Yes, I got the source but not the build

PR #234, opened 2026-06-22
Issue#233/custom filter term from request

### LINKED ISSUE #233: Inherit term setting on query loop blocks
### Background
We already have the `cataInheritQuery` feature that lets a Query Loop inherit the main query, which works well for a latest posts page. It would be useful to have a similar feature that filters a Query Loop by the current request's taxonomy term instead, without hardcoding the term into the saved block markup. Core's inherited query ignores offset, so the loop needs to stay non-inherited and have the term applied as a filter at render time.

### Details
- Base this off the existing `cataInheritQuery` feature at `block-editor/inherit-query/`. In `src/index.js`, add a `cataTermFromRequest` boolean attribute (default `false`) to `core/query`, and add an `InspectorControls` toggle ("Filter by current term").
- For the render logic, use the core `query_loop_block_query_vars` filter, which was added in WP 6.1. This only runs for non-inherited queries, which is what we want, since non-inherited loops honor offset. When the toggle is on, inject a `tax_query` for the current term.
- This feature will apply a `cata_term_from_request` filter and the host theme returns the term:

```php
$term = apply_filters( 'cata_term_from_request', null ); // host returns WP_Term|null

if ( $term instanceof WP_Term ) {
	$query['tax_query'][] = array(
		'taxonomy' => $term->taxonomy,
		'terms'    => array( $term->term_id ),
	);
}
```

### Notes
- We may need to expose `cataTermFromRequest` as block context so it's available in the filter, or we can read `$block->parsed_block['attrs']` directly.
- The editor preview doesn't run `query_loop_block_query_vars`, so the preview will show the latest posts even with the toggle on. The front end filters correctly.
- A concern Claude brought up: the `cataInheritQuery` feature needed a second `render_block_context` filter because `render_block_data` wasn't reliable on some hosts, and there's a comment about this in `inherit-query.php`. This feature uses a different hook, so we should confirm `query_loop_block_query_vars` actually runs on our hosts, and fall back to `render_block_context` or `pre_render_block` if it doesn't.

jessica-townsend:
### Related issues
- Closes #233 

### What was accomplished
- Moved `inherit-query` to new `query-filters` module
- Added a "Term From Request" filter to the query loop block

### How it was tested
- [x] Local

### How to test
- [ ] Add a query loop block to a page on a local site. Turn on "Filter by current term" under the Custom Query Filters settings panel. Ensure that the Query Type is set to Custom. Save the page.
- [ ] Verify on the frontend that the query loop block displays the latest posts.
- [ ] Add the following code to `functions.php` on the local site's active theme, updating `0` to a known category/tag ID:

```
function cata_test_term_from_request() {
	return get_term( 0 );
}
add_filter( 'cata_query_filters_term_from_request', 'cata_test_term_from_request' );
```
- [ ] Refresh the page with the query loop block. Verify that the posts displayed are consistent with the term ID added to the code above
- [ ] Custom inherit query filter still works as expected on a latest page


block-editor/query-filters/query-filters.php
```diff
@@ -0,0 +1,159 @@
...
+
+use WP_Block;
+use WP_Term;
+
+/**
+ * Enqueue Scripts
+ */
+function cata_query_filters_enqueue_scripts(): void {
+
+	if ( true === wp_is_block_theme() ) {
```
douglas-johnson: I think this is still useful in full site editing contexts. In the current use case we want to use offset in inherited taxonomy queries to piece together whatever layout we want and make it look like the query is flowing through the layout, even though its actually three queries with varying offsets.

The featured posts pattern we use on thought catalog that has three queries with varying offsets should still work if we set it to inherit the term and drop it into a full site editing template for categories or tags.

block-editor/query-filters/query-filters.php
```diff
@@ -0,0 +1,159 @@
...
+ * 
+ * @return array $query
+ */
+function cata_query_filters_term_from_request_query_vars( array $query, WP_Block $block, int $page ): array {
+
+	if ( true !== ( $block->context['cataTermFromRequest'] ?? false ) ) {
+		return $query;
+	}
+
+	$term = apply_filters( 'cata_query_filters_term_from_request', null );
```
douglas-johnson: Could this be `get_queried_object()` instead of null? I think that way it would work right away without the filter most of the time.

douglas-johnson (changes requested):
I think we want this to work for full site editing themes for ease of transition into FSE on any site where we end of using it.

block-editor/query-filters/query-filters.php
jessica-townsend: Removed, we had this for the custom inherit query filter so I had left it

block-editor/query-filters/query-filters.php
jessica-townsend: Good idea, updated

douglas-johnson (approved):

PR #236, opened 2026-06-29
Issue#235/image lightbox block

### LINKED ISSUE #235: Add image lightbox block for post images
Since more images are being added on posts, it would be great to have a feature where, when a user clicks on an image in the post, a lightbox modal appears to view all images in the post as a slider. We also want to be able to display an ad in the modal.

Reference here: https://nypost.com/2026/06/29/world-news/archaeologists-uncover-freeze-dried-potatoes-older-than-the-us-in-excellent-condition/

Some things to consider:
- WordPress already has an "Enalrge on click" feature, which can be enabled on a per-image basis
- It looks like NY Post uses a custom modal (no `wp-` classes in the inspector) and uses swiper for pagination
- Can this be a block that's added to the single post template?

jessica-townsend:
### Related Issues
- Closes #235 

### What Was Accomplished
- Added an image lightbox block which collects all images from a post's content and triggers a lightbox slider if an in-content image is clicked.

### How It Was Tested
- [x] Locally on parent theme
- [x] Locally on child block theme
- [x] Locally on child hybrid theme
- [x] Child block/hybrid theme develop (Thoughtnet) 

### How To Test
- On a FSE theme, add the image lightbox block to the single post template in the editor. Then, navigate to a post with multiple images in the content. 
  - [x] Clicking on an in-content image opens the modal
  - [x] The slider should navigate between all images in the content
  - [x] The slider should initially show the image that was clicked
  - [x] Slider works and looks appropriate on desktop and mobile
  - [x] Changing the color settings is reflected on the frontend
- On Collective World develop, navigate to a post with multiple images in the content.
  - [x] Slider functions the same as above

### Open Questions
- Once we get more information on ad placement, we can adjust the modal layout to include a space for an ad
- The slider doesn't include the featured image. We can adjust if we need to include that as well
- Currently the color settings aren't reflected in the editor. We can update that to have a "color preview" if we need to

### Deploy Steps
When applying this update to hybrid themes, we'll need to add this block to the PHP template using one of these options:

 Filtering `the_content`
```
add_filter( 'the_content', function ( $content ) {
	if ( is_singular() && in_the_loop() && is_main_query() ) {
		$content .= do_blocks( '<!-- wp:cata/image-lightbox /-->' );
	}
	return $content;
} );
```

Running `do_blocks` right after `the_content` is called (this is what's on Collective World develop now)
```
<?php echo do_blocks( '<!-- wp:cata/image-lightbox /-->' ); ?>
```


blocks/image-lightbox/image-lightbox.php
```diff
@@ -0,0 +1,311 @@
...
+			'style'               => $styles,
+		)
+	);
+
+	// Button contents; filter to swap the defaults for an SVG icon, etc.
+	$close_icon = apply_filters( 'cata_blocks_image_lightbox_close_icon', '×' );
+	$prev_icon  = apply_filters( 'cata_blocks_image_lightbox_prev_icon', '←' );
+	$next_icon  = apply_filters( 'cata_blocks_image_lightbox_next_icon', '→' );
+
+	ob_start();
```
douglas-johnson: Could this use a render.php file instead of a render_callback?

douglas-johnson:
I was not able to change the background or text color within the modal using an FSE theme. I also tried setting the modal to `color-scheme: dark only;` on Creepy Catalog which didn't seem to have any effect.

jessica-townsend:
Added updates for both of these on develop, @douglas-johnson I couldn't replicate not being able to change the bg/text color within the modal on FSE, it seemed to work for me.

The `color-scheme` setting didn't though, still testing that one to try and make it work.

douglas-johnson:
You are right, setting the colors did work, it was purely a color-scheme problem messing with the visualization of the colors ( picking light or dark )

douglas-johnson (approved):

PR #257, opened 2026-07-06
Moments Hero → Multi-Block Decomposition

douglas-johnson:
**Branch:** [`feature/moments-hero-multi-block`](https://github.com/thoughtis/cata-blocks/tree/feature/moments-hero-multi-block) (continues from `feature/moments-hero-block`, PR #255)
**Block docs:** [`blocks/art-direction/README.md`](https://github.com/thoughtis/cata-blocks/tree/feature/moments-hero-multi-block/blocks/art-direction)

This branch reworks PR #255. The `cata/moments-hero` block worked, but it was a monolith: it owned post targeting, image selection, layout, *and* the title/date markup — and its CSS reached into theme presets (`--wp--preset--font-family--sans-serif-condensed`, `--wp--preset--color--fg-min`, the type scale), coupling the plugin to one theme in a way the block editor normally lets us avoid. The branch **removes `cata/moments-hero`** and replaces it with three generic pieces that compose in a Query Loop, so typography and color go back to the theme.

### 1. Slot targeting — mostly core, plus one query filter

"Slot N = Nth newest pinned post" turned out to need almost no custom code. Core's Query Loop already does it:

- **"Only pinned posts"** (`query.sticky = "only"`) + **order by date desc** + **1 per page** + **offset N−1** = the Nth newest pinned post.
- When nothing is pinned, core resolves the pin list to `post__in => [0]`, so the loop renders **nothing** — exactly the old slot 2/3 behavior, for free.

The only custom part is the old slot 1 guarantee ("the hero is never blank"). That's now a Custom Query Filter on `core/query`, alongside the existing Inherit Query / Filter by current term toggles in `block-editor/query-filters/`:

- **`cataPinnedFallback`** (toggle: *Fallback when nothing is pinned*) — when the loop shows only pinned posts and no published pinned post exists, the query swaps to the latest post that has a featured image.
- **`cataPinnedFallbackCategories`** (optional ID list) — scopes that fallback query, replacing the old `fallbackCategories` attribute.

Implementation follows the `cataTermFromRequest` pattern exactly: attributes + context via `block_type_metadata`, behavior via `query_loop_block_query_vars`.

### 2. `cata/art-direction` — the layout block

New block at `blocks/art-direction/`. Server-rendered from `postId` context (works inside any post template), with a single attribute:

| `layout` | Renders | Degrades to |
| --- | --- | --- |
| `auto` (default) | 2×2 collage when the post carries ≥4 good same-orientation images | Single |
| `single` | Featured image, natural ratio | Nothing when no featured image |
| `diptych` | Featured + best second image (portrait preferred) | Single |
| `collage` | 2×2 grid of four good same-orientation images | Single |

The old slot-number branching ("slot 2 gets the diptych") became per-instance configuration: the slot 2 loop's template sets `layout: "diptych"`, slots 1/3 use `auto`. The "good image" rules (in-content attachments with alt text, no PNGs, no featured-image duplicates) moved verbatim from `moments-hero/includes/selection.php` to `art-direction/includes/selection.php`.

When nothing is renderable the block outputs nothing, so the surrounding title/date form a text-only card — the old art-less-pinned-post behavior, now emergent instead of special-cased.

The block's CSS is **structural only**: diptych flex, collage grid, 4px gaps, `object-position: center top` crop anchoring. No theme tokens.

### 3. `cata_art_direction` REST field — editor preview data

`art-direction/includes/rest.php` registers a REST field on posts exposing the featured image, the good images, **the exact images the diptych and collage would use**, and which layout `auto` resolves to. It shares `selection.php` with the render template, so the editor preview and the server render cannot drift.

The block's `edit.js` uses it to show real-image previews of the chosen layout for the actual post in the loop, plus a note like *"Auto renders this post as: Collage"* or *"Not enough usable images — renders as Single."*

The field is computed for **`context=edit` requests only** (which is what the editor sends); public/anonymous API responses skip the content scan entirely and return null.

## The hero as a theme pattern

Title and freshness/date now come from `core/post-title` and the existing `cata/fresh-post-dot` (wrapping `core/post-date` in `human-diff` format — the old block's meta line was a re-implementation of it, same hour-rounding and all). The full three-loop pattern markup, ready for the theme, is in the [block README](https://github.com/thoughtis/cata-blocks/tree/feature/moments-hero-multi-block/blocks/art-direction) — shape:

```html
<!-- wp:query {"query":{"perPage":1,"offset":0,"sticky":"only","order":"desc","orderBy":"date","postType":"post","inherit":false},"cataPinnedFallback":true,"cataPinnedFallbackCategories":[...]} -->
  <!-- wp:post-template -->
    <!-- wp:cata/art-direction {"layout":"auto"} /-->
    <!-- wp:post-title {"isLink":true} /-->
    <!-- wp:cata/fresh-post-dot --><!-- wp:post-date {"format":"human-diff"} /--><!-- /wp:cata/fresh-post-dot -->
  <!-- /wp:post-template -->
<!-- /wp:query -->
```

…then the same with `offset: 1` + `layout: "diptych"`, and `offset: 2` + `layout: "auto"`. Type scale, fonts, and colors are styled at the pattern/theme.json level.

## Verification

Smoke-tested in WordPress Playground (WP 7.0, PHP 8.3, plugin mounted) with engineered fixtures — 19/19 assertions:

- Selection: auto→collage (4 imgs, PNG + empty-alt decoys skipped), diptych picks the portrait, featured-only→single, diptych degrades to single.
- REST: field present with correct picks in edit context, null in view context.
- Rendering: three sticky-slot loops render A/B/C as collage/diptych/single; with nothing pinned, the fallback shows the latest-with-thumbnail post site-wide, the category-scoped fallback shows the scoped post, and non-fallback loops render no post.

`wp-scripts` builds are committed (`query-filters` and `art-direction`); all PHP passes `php -l`. PHPCS couldn't run locally — the vendored WPCS fatals under PHP 8.4 before checking anything (pre-existing toolchain issue, worth confirming in CI).

## Notes for whoever picks this up

- **Everything is one commit** on the branch (`replace moments hero block with art direction block…`), which reads as a rename+rewrite of the PR #255 diff. Suggest retargeting PR #255 to this branch or opening a fresh PR.
- **Rollout order changes slightly** from the PR #255 plan: the theme needs the pattern (and any pattern-level styling) *before* the homepage swap, since the block no longer carries title/meta styles. Groomer retirement and removal of the interim `tc-moments-*` Additional CSS rules are unchanged.
- **Editor niceties left open:** the fallback-categories control is a comma-separated-ID text field (matching the minimal query-filters panel style); a category token/search field would be friendlier if editors configure it by hand rather than via pattern markup. The pinned fallback also doesn't preview in the editor (it's a server-side query swap) — the editor just shows the loop's empty state.
- **Semantic choice to know about:** `auto` never picks the diptych; the diptych is always an explicit per-instance choice. This matches the old slot rules (only slot 2 ever got the diptych) while keeping `auto` predictable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


douglas-johnson:
Merged in #261

